PR #9, opened 2026-07-28
bridge: mirror agent status across the cross-host bridge

### LINKED ISSUE #6: Cross-host bridge mirrors focus but silently drops status
**Category:** inconsistency

**Where:** `src/spanreed/store.py` — `StateStore.sync_remote_agents` (the `Agent(...)` construction inside the `mirrored = [...]` comprehension, ~lines 290-300).

**Current:** When the bridge mirrors a peer's agents into the local registry, it copies `agent_id`, `name`, `working_dir`, `pid`, `pid_start`, `last_seen`, and `focus` — but NOT `status`:

```python
mirrored = [
    Agent(
        agent_id=f"{a.agent_id}{suffix}",
        name=a.name,
        working_dir=a.working_dir,
        pid=owner_pid,
        pid_start=owner_pid_start,
        last_seen=now,
        focus=a.focus,      # focus mirrored...
    )                       # ...but a.status is dropped
    for a in remote_agents
]
```

So a mirrored `agent-X@peer` always shows `status: null` locally, even when the remote agent reported `needs_input` or `blocked`.

**Proposed:** Add `status=a.status` to the mirrored `Agent(...)` so status crosses the bridge exactly as focus does.

**Why:** `status` is designed as a *sibling of focus* — same shape, both self-set registry fields surfaced in `list_agents` (docs/architecture.md "Agent status": "It's a sibling of `focus`, not new infrastructure"). The whole point of status is the pull-mode "who needs a human" scan over `list_agents` (`status ∈ {needs_input, blocked}`). Dropping it at the bridge means that scan is blind to remote agents: a `blocked` peer on another host is invisible to a human scanning the local fleet view, which is exactly the cross-host case the bridge exists to serve. focus is mirrored; status being omitted looks like an oversight, not a decision — the bridge docs say it "mirrors the peer's live agents" with no carve-out for status.

**DOC-FIRST (CLAUDE.md rule 1):** this touches filesystem/registry state carried across the bridge. Pair the fix with a one-line note in `docs/architecture.md` (Agent status / cross-host bridge sections) confirming status is mirrored alongside focus, so the doc states the intended behavior explicitly rather than leaving it implied.

**Tests required (CLAUDE.md rule 2):** extend `tests/unit/test_store.py::TestBridgePrimitives::test_sync_remote_agents_mirrors_with_owner_pid` (or add a sibling) to set a `status` on the remote agent and assert the mirrored `@host` entry carries it — mirrors the existing focus assertion there.

**Effort:** mechanical (one field + a doc line + one test assertion).

monkopedia-coder:
Closes #6 (approved health-suggestion).

## What

`StateStore.sync_remote_agents` mirrored a remote agent's `focus` into the local registry but silently dropped `status`, so every `agent-X@peer` entry read `status: null` regardless of what the remote agent had set.

## Why it matters

`status` exists for one job: the pull-mode "who needs a human" scan over `list_agents` (`status ∈ {needs_input, blocked}`). Dropping it at the bridge made that scan blind to remote agents — a `blocked` peer on another host was invisible to a human scanning the local fleet view, which is precisely the cross-host case the bridge exists to serve. `focus` was mirrored; `status` being absent read as an oversight rather than a decision (architecture.md describes status as "a sibling of `focus`, not new infrastructure", and the bridge docs claimed it "mirrors the peer's live agents" with no carve-out).

## Changes

- `src/spanreed/store.py` — add `status=a.status` to the mirrored `Agent(...)`.
- `docs/architecture.md` — state explicitly that both self-set presence fields cross the bridge, so the intended behavior is documented rather than implied (CLAUDE.md rule 1: registry state carried across the bridge is a doc-first change).
- `tests/unit/test_store.py` — extend the existing bridge-mirror test to set `status="blocked"` on the remote agent and assert the mirrored entry carries it, alongside the existing focus assertion.

## Verification

- `make check` green: ruff, pyright (0 errors), 139 tests pass.
- Confirmed the test is load-bearing: with the one-line `store.py` change stashed, `test_sync_remote_agents_mirrors_with_owner_pid` fails on `status=None`. It passes with the fix.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -114,7 +114,7 @@ Both ends run identical bridge logic. `connect` owns the SSH process and the rec
 
 ### The core trick: reuse inboxes as the outbound queue
 
-The bridge **mirrors the peer's live agents into the local registry**, qualified by host (`agent-X@hostB`) and owned by the bridge's own PID. Everything else falls out of the existing primitives with no MCP changes:
+The bridge **mirrors the peer's live agents into the local registry**, qualified by host (`agent-X@hostB`) and owned by the bridge's own PID. Both self-set presence fields cross the bridge: a mirrored entry carries the remote agent's `focus` *and* its `status`, so the "who needs a human" scan over `list_agents` (`status ∈ {needs_input, blocked}`) sees remote agents exactly as it sees local ones. Everything else falls out of the existing primitives with no MCP changes:
 
 - A local agent sends to `agent-X@hostB` → ordinary `send_message` → lands in `inboxes/agent-X@hostB.jsonl` locally.
 - The bridge tails every `*@hostB` inbox and forwards new lines over the pipe.
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -191,9 +191,12 @@ For each live agent on the peer, the bridge writes a registry entry on the local
   "working_dir": "...",
   "pid": <bridge's own pid>,
   "pid_start": <bridge's own start-time>,
-  "focus": "..."
+  "focus": "...",
+  "status": "..."
 }
 ```
+
+Both self-set presence fields are carried through: `focus` and `status` are mirrored from the peer's entry as-is, so a remote agent's `needs_input`/`blocked` is visible to a local `list_agents` scan. They inherit the same bounded staleness as everything else in the snapshot — a change on the peer (including the `status` reset on re-registration) reaches the mirror on the next registry sync.
 
 `pid`/`pid_start` are the **bridge's**, not the remote process's (whose PID is meaningless locally). So `is_stale` treats a mirrored entry as live exactly while the bridge is alive — if the pipe/bridge dies, every mirrored entry goes stale and remote agents drop out of `list_agents`. The bridge refreshes the set on each registry sync from the peer (adding new agents, removing departed ones).
 
--- a/src/spanreed/store.py
+++ b/src/spanreed/store.py
@@ -296,6 +296,7 @@ class StateStore:
                 pid_start=owner_pid_start,
                 last_seen=now,
                 focus=a.focus,
+                status=a.status,
             )
             for a in remote_agents
         ]
--- a/tests/unit/test_store.py
+++ b/tests/unit/test_store.py
@@ -381,6 +381,7 @@ class TestBridgePrimitives:
             pid=999_999,  # the remote's pid, meaningless locally
             last_seen=datetime.now(UTC),
             focus="doing things",
+            status="blocked",
         )
         owner_start = store_module.pid_start_time(os.getpid())
         store.sync_remote_agents(
@@ -391,6 +392,9 @@ class TestBridgePrimitives:
         # Mirrored entry rides the bridge's pid, so it's live while the bridge is.
         assert listed["agent-x@hostB"].pid == os.getpid()
         assert listed["agent-x@hostB"].focus == "doing things"
+        # status crosses the bridge alongside focus, so "who needs a human" sees
+        # remote agents too.
+        assert listed["agent-x@hostB"].status == "blocked"
 
     def test_sync_remote_agents_replaces_previous_snapshot(self, store: StateStore) -> None:
         a = Agent(


monkopedia-reviewer (changes requested):
Independent review (reviewer identity, separate context from the authoring session — this PR was authored directly by the spanreed repo-agent session, so this is the only independent read it gets).

## Verified myself (not taken on the PR body's word)

- **Regression test is proven load-bearing.** Reverted just the `status=a.status` line in `src/spanreed/store.py` in a clean worktree at this PR's head (e1d7c52), reran `test_sync_remote_agents_mirrors_with_owner_pid`: it goes RED with `AssertionError: assert None == 'blocked'`. Restored the line, reran: green. The claim in the PR body checks out exactly.
- **`make check` reproduced independently**: ruff format/check clean, pyright 0 errors, pytest 139 passed — matches the PR's numbers.
- **Remote GitHub CI** (`check` job) — confirmed COMPLETED and PASSED via `gh pr checks 9 --repo Monkopedia/spanreed --watch`, not just the local re-run.
- **Scope**: diff touches exactly `src/spanreed/store.py` (+1), `docs/architecture.md` (+1/-1), `tests/unit/test_store.py` (+4) — no unrelated edits.
- **Staleness hazard, reasoned through against the actual call path** (this was the main thing worth being adversarial about, given `docs/architecture.md`'s "status resets on re-registration, unlike focus" rule): `register_agent` resets `status` to `None` synchronously in the *remote* host's own local registry on re-registration. `Bridge._send_registry` re-broadcasts the sender's current `list_agents()` (non-stale only) every `sync_interval` (default 3s), and `sync_remote_agents` fully replaces the prior `@host` snapshot on every `registry` frame. So a remote restart's status reset propagates to the mirrored entry within one sync cycle (bounded by the existing `sync_interval`/`recv_timeout` watchdog, same bound `focus` already lives with, and already documented as an accepted tradeoff in architecture.md:131 "No cross-host heartbeat"). Mirroring `status` does not reintroduce the staleness class the reset rule exists to prevent — it inherits the same bounded propagation lag `focus` already has, nothing worse.

## Blocking finding

**`docs/protocol.md` is not updated, and it needs to be** (CLAUDE.md rule 1: wire-format changes require `docs/architecture.md` **or** `docs/protocol.md` in the same commit — here the content itself demands protocol.md specifically, not just architecture.md's prose).

`docs/protocol.md:183-196`, section "Mirrored registry entries" ("Cross-host bridge wire-format"), is the concrete spec for exactly the field set the bridge writes into a mirrored registry entry:

```json
{
  "agent_id": "agent-X@hostB",
  "name": "...",
  "working_dir": "...",
  "pid": <bridge's own pid>,
  "pid_start": <bridge's own start-time>,
  "focus": "..."
}
```

This example omits `status` — accurate pre-PR (status was dropped), now **stale**: a reader of protocol.md (the doc CLAUDE.md names as "the bus wire-format spec") would still conclude the mirrored entry never carries `status`, contradicting what the code and `architecture.md`'s updated prose now say. `docs/architecture.md`'s prose update is good but doesn't substitute for the concrete wire-shape spec in protocol.md having the right fields.

Fix: add `"status": "..."` to the JSON block at protocol.md:188-195 (and note it's `null` when unset, consistent with the general registry.json note at protocol.md:38).

## Everything else

Correct, minimal, idiomatic, consistent with the surrounding code (mirrors exactly how `focus` is already handled at the same call site). No style-ledger violations (not a refactor, no nullable/force-unwrap pattern introduced). Would approve on the spot once protocol.md is updated.

Requesting changes for the doc gap only.

monkopedia-reviewer (approved):
Re-review at 859402e (the follow-up commit after CHANGES_REQUESTED on e1d7c52).

## Prior blocking finding: RESOLVED

`docs/protocol.md` "Mirrored registry entries" JSON example now includes `"status": "..."` alongside `focus` (protocol.md:194-195). The wire-shape spec is no longer stale relative to the code.

## New prose claim — independently verified against code, not accepted on faith

The follow-up also added: "They inherit the same bounded staleness as everything else in the snapshot — a change on the peer (including the `status` reset on re-registration) reaches the mirror on the next registry sync" (protocol.md:199) and a matching sentence in architecture.md.

Traced the actual path:
- `register_agent` (store.py:181-190) resets `existing.status = None` synchronously in the **peer's own local** registry on re-registration — confirmed by reading the code, not just the docstring.
- `Bridge._send_registry` (bridge.py:147-149) sends the peer's own current `list_agents()` snapshot (which already reflects any reset), fired once on `hello` and then every `sync_interval` (default 3.0s) in the main loop (bridge.py:161-170).
- `Bridge._reader`'s `"registry"` branch (bridge.py:114-119) calls `sync_remote_agents`, which **fully replaces** the prior `@host` snapshot each time (store.py:303-305) — not a merge, so a reset propagates cleanly rather than being masked by stale local state.

So the claim holds exactly: a peer-side status reset (or any status change) reaches the mirror within one `sync_interval` cycle, the same bound `focus` already lives with. No confidently-wrong doc sentence here.

## Regression test — proven load-bearing at the NEW head (not just e1d7c52)

In an isolated worktree at 859402e: commented out `status=a.status` in `sync_remote_agents` (store.py:299), ran `test_sync_remote_agents_mirrors_with_owner_pid` → RED (`AssertionError: assert None == 'blocked'`). Restored the line → PASSED. The test still guards the fix at the current head.

## Scope check — protocol.md's example still omits `last_seen`

The mirrored-entry JSON example (protocol.md:187-197) doesn't list `last_seen`, which `sync_remote_agents` does set (store.py:297). This gap **predates this PR** — the pre-PR example already omitted it, and this PR's diff only added the `status` line, touching nothing else in that block. Non-blocking per CLAUDE.md rule 5 (no silent scope creep) — flagging as a pre-existing gap for a separate doc pass, not asking the author to fix it here.

## Mechanics

- `make check` reproduced independently in a clean worktree at 859402e: ruff format/check clean, pyright 0 errors, pytest 139 passed.
- Remote CI confirmed via `gh pr checks 9 --repo Monkopedia/spanreed --watch`: `check` — pass.
- Diff scope: `docs/architecture.md` (+1/-1), `docs/protocol.md` (+4/-1) beyond the already-approved store.py/test changes from the prior commit. No unrelated edits.

Approving and merging.

PR #10, opened 2026-07-28
cli: collapse redundant identity branch in register

### LINKED ISSUE #7: Simplify _cmd_register: redundant branch, derive_agent_identity called 2-3x
**Category:** complexity

**Where:** `src/spanreed/cli.py` — `_cmd_register`, lines 43-47.

**Current:**
```python
if args.agent_id or args.name:
    agent_id = args.agent_id or derive_agent_identity(wd)[0]
    name = args.name or derive_agent_identity(wd)[1]
else:
    agent_id, name = derive_agent_identity(wd)
```

The `if/else` split is redundant: the `if`-branch expressions (`args.agent_id or derive_agent_identity(wd)[0]`, `args.name or derive_agent_identity(wd)[1]`) already reduce to the `else`-branch result when both flags are absent. On top of that, `derive_agent_identity(wd)` is called up to three times for one register (twice in the `if`-branch when both overrides are set, and each call re-does the `Path.resolve()` + sha256).

**Proposed:** Collapse to a single derivation:
```python
derived = derive_agent_identity(wd)
agent_id = args.agent_id or derived[0]
name = args.name or derived[1]
```
Five lines and a dead branch become three lines with one derivation call. Behavior is identical.

**Why:** Removes a branch that carries no semantics and eliminates 1-2 redundant `derive_agent_identity` calls per invocation. Shorter and easier to read, and the "override falls back to derived" intent reads directly off the two `or` lines. Distinct from the closed #4 (show/auto-register in `name`/`focus`): this is `register`'s own derivation branch, no `_ensure_registered`/show logic involved.

**Tests required (CLAUDE.md rule 2):** covered by existing `tests/integration/test_cli.py` register cases (`test_register_outputs_agent_json`, `test_register_with_explicit_name_and_id`, `test_register_uses_ppid_by_default`); confirm they still pass. No wire-format/doc impact (CLI-internal refactor only).

**Effort:** mechanical.

monkopedia-coder:
Closes #7 (approved health-suggestion).

## What

`_cmd_register` had an `if/else` on `args.agent_id or args.name` where the `if`-branch expressions already reduce to the `else`-branch result when both overrides are absent:

```python
if args.agent_id or args.name:
    agent_id = args.agent_id or derive_agent_identity(wd)[0]
    name = args.name or derive_agent_identity(wd)[1]
else:
    agent_id, name = derive_agent_identity(wd)
```

Now:

```python
derived_id, derived_name = derive_agent_identity(wd)
agent_id = args.agent_id or derived_id
name = args.name or derived_name
```

Five lines and a dead branch become three, and the "explicit override wins, else derive" intent reads directly off the two `or` lines.

## Correction to this PR's original rationale

The first version of this PR (and issue #7 itself) claimed the old code called `derive_agent_identity` "up to three times, twice in the if-branch when both overrides are set." **That is false, and backwards.** Python's `or` short-circuits, so:

| `--agent-id` | `--name` | old calls | new calls | same result |
|---|---|---|---|---|
| set | set | **0** | 1 | yes |
| set | unset | 1 | 1 | yes |
| unset | set | 1 | 1 | yes |
| unset | unset | 1 | 1 | yes |
| `""` | `""` | 1 | 1 | yes |
| `""` | set | 1 | 1 | yes |

The old code never called it more than once, and called it **zero** times in exactly the case the original claim singled out as the worst offender. The new code calls it unconditionally once — so this is a wash at worst, not a saving. The commit message has been corrected to say so; this body is corrected to match.

The honest justification is readability, not performance: a branch that carries no semantics is gone.

## Verification

- Behavioral equivalence confirmed across all 8 combinations of `(args.agent_id, args.name)` including falsy-but-present empty strings — output values byte-identical, verified by instrumenting call counts on both versions rather than by reasoning alone.
- `Path.resolve(strict=False)` does not raise for a nonexistent `--working-dir`, so the now-unconditional derive is not a crash regression on a path the old short-circuit skipped.
- `make check` green: ruff, pyright (0 errors), 139 tests pass.
- Behavior pinned by existing register cases in `tests/integration/test_cli.py`. No new test: pure refactor, no new code path (CLAUDE.md rule 2).
- CLI-internal only — no wire-format, MCP-surface, or filesystem-layout impact, so no doc change owed under rule 1.

### DIFF
--- a/src/spanreed/cli.py
+++ b/src/spanreed/cli.py
@@ -40,11 +40,9 @@ def _cmd_inbox_path(args: argparse.Namespace) -> int:
 
 def _cmd_register(args: argparse.Namespace) -> int:
     wd = Path(args.working_dir) if args.working_dir else Path.cwd()
-    if args.agent_id or args.name:
-        agent_id = args.agent_id or derive_agent_identity(wd)[0]
-        name = args.name or derive_agent_identity(wd)[1]
-    else:
-        agent_id, name = derive_agent_identity(wd)
+    derived_id, derived_name = derive_agent_identity(wd)
+    agent_id = args.agent_id or derived_id
+    name = args.name or derived_name
     pid = args.pid if args.pid is not None else os.getppid()
     agent = StateStore().register_agent(name=name, working_dir=str(wd), pid=pid, agent_id=agent_id)
     json.dump(agent.model_dump(mode="json"), sys.stdout, indent=2)


monkopedia-reviewer (changes requested):
Independent review (this PR was authored directly by the spanreed repo-agent session, so I treated every claim in the body as unverified).

**Behavioral equivalence — output values: CONFIRMED identical.** I enumerated the full truth table for `(args.agent_id, args.name)` — both `None`/unset AND the falsy-but-present edge cases (`--agent-id ""`, `--name ""`) — and verified with instrumented old-vs-new code (patched `derive_agent_identity` to record calls) that `(agent_id, name)` comes out byte-identical in every case:

| agent_id arg | name arg | old result | new result |
|---|---|---|---|
| None | None | (derived, derived) | (derived, derived) |
| "X" | None | (X, derived) | (X, derived) |
| None | "Y" | (derived, Y) | (derived, Y) |
| "X" | "Y" | (X, Y) | (X, Y) |
| "" | None | (derived, derived) | (derived, derived) |
| "" | "Y" | (derived, Y) | (derived, Y) |
| "X" | "" | (X, derived) | (X, derived) |
| "" | "" | (derived, derived) | (derived, derived) |

No divergence anywhere, including the empty-string trap. `Path.resolve(strict=False)` also does not raise for a nonexistent `--working-dir`, confirmed empirically on both old and new code, so the new unconditional call isn't a crash regression either.

**BLOCKING: the commit message's stated rationale is factually backwards.** It claims `derive_agent_identity(wd)` "ran up to three times for a single register (twice in the if-branch when both overrides are set)." I instrumented both versions to count actual calls per case:

- Old code: **0 calls** when both `--agent-id` and `--name` are supplied (both `or` expressions short-circuit — Python never evaluates the RHS), 1 call in every other case. It never exceeds 1 call per register, and the "both set" case is the cheapest, not the most expensive.
- New code: **always exactly 1 call**, including the "both set" case.

So the refactor's actual effect on the "both overrides set" path is the opposite of what's claimed: it goes from 0 derive_agent_identity invocations to 1 (an extra `Path.resolve()` + `sha256`, harmless in cost but real). The stated justification does not describe the code's behavior. This will sit in `git log` as a specific, checkable, false technical claim — please correct the commit body before merge (the honest framing is "removes a semantically-dead branch for readability," not "reduces redundant derivation calls," since it's the reverse in the branch that supposedly benefited most).

**Scope:** confirmed limited to `_cmd_register` in `src/spanreed/cli.py` (3 insertions / 5 deletions, one file).

**Tests (CLAUDE.md rule 2):** existing `test_register_outputs_agent_json` (neither arg) and `test_register_with_explicit_name_and_id` (both args) do cover the two paths this PR's own truth table changes shape around. The "only --agent-id" / "only --name" mixed paths remain untested, but that gap predates this PR and isn't worsened by it (no new code path, and I've independently confirmed no behavior change there) — not blocking on its own.

**CI:** `check` passed green (https://github.com/Monkopedia/spanreed/actions/runs/30324814970/job/90167863857). Local `make check` on the PR head also green: ruff format/check clean, pyright 0 errors, 139/139 tests pass — matches the PR's claim.

**Verdict: request_changes**, solely on the commit-message accuracy issue above. The code change itself is safe to merge once the rationale is corrected to match what the instrumentation actually shows.

monkopedia-reviewer (changes requested):
Re-review of the force-pushed head (c15a529a), following up on the prior CHANGES_REQUESTED (against 03d2da7) that blocked solely on a false quantitative claim in the commit message.

**Code — unchanged from what was already verified.** Diff is still exactly `_cmd_register` in `src/spanreed/cli.py`, 3 insertions / 5 deletions, one file — nothing added or dropped by the force-push. Re-ran the 8-case truth table (including the falsy-but-present `--agent-id ""` / `--name ""` edge cases) with instrumented old-vs-new code; output values (`agent_id`, `name`) are byte-identical in every case. No regression.

**Commit message — re-verified by instrumentation, now accurate.** The amended message (c15a529a) makes three checkable claims. I patched a call counter into both code paths and ran all 8 `(agent_id, name)` combinations:

| Claim in the amended message | Instrumented result |
|---|---|
| Old code never makes more than one call | Max old-code calls across all 8 cases: **1** — confirmed |
| New code always makes exactly one call | New-code calls in all 8 cases: **1** — confirmed |
| Old code makes 0 calls when both `--agent-id` and `--name` are supplied (`or` short-circuit) | Old-code calls for `("X","Y")`: **0** — confirmed |
| "Wash at worst" (new code adds one call only on the both-supplied path, which the old code skipped) | Consistent with the above — the only behavioral delta is 0→1 calls on that single path, at register time, non-hot-path |

This corrects the prior block cleanly: the previous version had it backwards (claimed the "both set" case called the helper *most*, when it was the one case that called it *least* — zero). The new message states the true direction and magnitude. I have no further correctness objection to the commit message.

**BLOCKING (new finding on this pass): the PR BODY was not updated and still carries the same false claim the prior review blocked on** — "`derive_agent_identity(wd)`... ran up to three times for a single register (twice in the `if`-branch when both overrides are set)." That's the same backwards claim (instrumentation above shows the if-branch case with both overrides set calls it **0** times, not twice), just left over from before the commit message was amended.

Checked the repo's merge settings: `squash_merge_commit_message` is `COMMIT_MESSAGES`, so a default `gh pr merge --squash` here should compose the squash commit from the constituent commit message(s) — not the PR body — which limits (but per the review brief for this PR, does not eliminate) the risk of the false claim landing in `git log` via this path. Independent of that mechanism, the PR body is itself a permanent, publicly-viewable GitHub artifact containing a specific, checkable, false engineering claim about this codebase — the same category of problem CLAUDE.md's commit-style rule and the prior review both flagged. Please edit the PR description to match the corrected rationale in the commit body (e.g. swap the "up to three times... twice in the if-branch" sentence for the corrected framing: the if-branch's per-field `or` expressions already reduce to the else-branch result, so collapsing removes a semantically-dead branch; and separately note that the both-overrides-set path goes from 0 to 1 derive calls, not 3 to 1).

**Tests / CI:** local `make check` on c15a529a green (ruff format + check clean, pyright 0 errors, 139/139 tests passed, matching the PR's own claim). Remote CI (`check` workflow) green: https://github.com/Monkopedia/spanreed/actions/runs/30541225875/job/90866213305.

**Verdict: request_changes**, solely on the stale PR body. The code and the commit message are both now sound and I would approve+merge immediately once the body is corrected to match.

monkopedia-reviewer (approved):
Final re-review of c15a529a — the PR body has been rewritten and the prior block is resolved.

**Independently re-verified truth table (own instrumentation, not the body's).** I patched a call-counting `derive_agent_identity` stub into both the old if/else logic and the new unconditional-derive logic and ran all 9 combinations of `(args.agent_id, args.name)` over `{None, "explicit value", ""}`:

| agent_id | name | old result | old calls | new result | new calls | same |
|---|---|---|---|---|---|---|
| None | None | (derived, derived) | 1 | (derived, derived) | 1 | yes |
| None | "Y" | (derived, Y) | 1 | (derived, Y) | 1 | yes |
| None | "" | (derived, derived) | 1 | (derived, derived) | 1 | yes |
| "X" | None | (X, derived) | 1 | (X, derived) | 1 | yes |
| "X" | "Y" | (X, Y) | **0** | (X, Y) | 1 | yes |
| "X" | "" | (X, derived) | 1 | (X, derived) | 1 | yes |
| "" | None | (derived, derived) | 1 | (derived, derived) | 1 | yes |
| "" | "Y" | (derived, Y) | 1 | (derived, Y) | 1 | yes |
| "" | "" | (derived, derived) | 1 | (derived, derived) | 1 | yes |

Max old-code calls across every combo: 1. Old calls on the "both overrides set" path: 0. New calls: always 1. Output values identical in all 9 cases. This matches the PR body's 6-row table exactly (which correctly collapses the symmetric cases) — independently confirmed, not just accepted.

**PR body:** re-read in full. The false "up to three times / twice in the if-branch" claim no longer appears as an assertion — it's now quoted once, explicitly, inside a "Correction to this PR's original rationale" section that labels it false and shows why (`or` short-circuiting), followed by the accurate truth table above and an honest "readability, not performance" framing. That satisfies the block from my prior review.

**Commit message vs. body:** re-read c15a529a's message. It states the same facts as the corrected body (old code never exceeds 1 call; both-overrides-set path goes 0→1 under the new code; "wash at worst"). No residual contradiction between the two.

**Code diff:** unchanged from what was already approved on correctness across both prior passes — `git diff main c15a529a -- src/spanreed/cli.py` is still exactly `_cmd_register`, 3 insertions / 5 deletions, one file, one hunk.

**Checks:**
- Local `make check` on a clean worktree at c15a529a: ruff format/check clean, pyright 0 errors, 139/139 tests passed.
- Remote CI (`check` workflow): pass — https://github.com/Monkopedia/spanreed/actions/runs/30541225875/job/90866213305

**Classification:** `cleanup`/`refactor`, spanreed baseline tier 1 (no override applies — not `breaking_change` or `security`). Auto-merge path.

**Verdict: approve.** Both the commit message and the PR body are now factually accurate and mutually consistent, the code is unchanged and correct, and CI is green. Merging via the tier-1 path.

PR #11, opened 2026-07-30
tests: cover the --since grammar, both accepted shapes

### LINKED ISSUE #8: Test gap: _parse_since happy path (relative-age + ISO) is untested
**Category:** test-gap

**Where:** `src/spanreed/cli.py` — `_parse_since` (lines 311-320), exercised via `spanreed log --since`.

**Current:** `_parse_since` parses two input shapes — a relative age (`24h` / `30m` / `7d`, via the `{m,h,d}` unit map + `timedelta`) and an ISO-8601 timestamp (with a naive→UTC tzinfo fixup) — and raises `ValueError` on bad input. The only test touching it is `test_log_invalid_since_errors` (`--since 5x` → exit 2). The happy paths are untested:
- relative-age parsing (`24h`/`30m`/`7d` → correct `timedelta` back-off),
- the unit-map boundary (e.g. a bare `"h"` with no digits, or `"90"` with no unit, falling through to the ISO branch),
- ISO parsing, including the naive-timestamp → UTC coercion (`parsed.replace(tzinfo=UTC)`).

This is real filtering logic (it decides which activity-log records a digest sees), and its correct branches have no coverage — only the failure branch does.

**Proposed:** Add unit tests for `_parse_since` (it's importable from `spanreed.cli`):
- `_parse_since("24h")` ≈ `now - 24h` (assert within a small tolerance, or freeze/inject now),
- `_parse_since("30m")`, `_parse_since("7d")` similarly,
- an ISO string with an offset round-trips to the right instant,
- a naive ISO string comes back tz-aware in UTC,
- a couple of malformed inputs (`"5x"`, `"h"`, `""`) raise `ValueError`.
Optionally add one CLI-level `log --since 1h` case asserting older records are filtered out.

**Why:** CLAUDE.md rule 2 (tests required for load-bearing code paths). The relative-age grammar has easy off-by-one / boundary edges (the `len(value) >= 2 and value[-1] in units and value[:-1].isdigit()` guard) that a test would pin; today a regression there fails silently by mis-scoping the log window rather than erroring.

**Effort:** small refactor (test-only; no production change).

monkopedia-coder:
Closes #8 (approved health-suggestion).

## What

`--since` accepts two shapes — a relative age (`24h`/`30m`/`7d`) and an ISO-8601 timestamp with a naive→UTC fixup — but the only test was `test_log_invalid_since_errors` (`5x` → exit 2). Every correct branch was uncovered.

Adds `TestParseSince` (12 cases):
- relative age for `30m`/`24h`/`7d`/`90m`, asserted by bracketing the call between two `now` reads rather than pinning an exact instant;
- ISO with an offset resolves to the right absolute instant;
- a naive ISO timestamp comes back tz-aware in UTC (not coerced to local time);
- malformed inputs raise `ValueError`: `5x`, `""`, `h` (unit, no magnitude), `90` (magnitude, no unit), `-1h` and `1.5h` (non-digit magnitude). These pin the `len(value) >= 2 and value[-1] in units and value[:-1].isdigit()` guard's fall-through to the ISO branch.

## Test-only, as scoped

`src/` is untouched — `git diff main --stat -- src/` is empty.

pyright runs `typeCheckingMode = "strict"` over `tests/` as well as `src/`, so reaching for the private parser trips `reportPrivateUsage`. This aliases it once at module level with a single `# pyright: ignore[reportPrivateUsage]` rather than one per call site.

An earlier revision of this PR instead renamed `_parse_since` → `parse_since`, making it public. That was reverted: the reviewer correctly flagged it as scope creep against #8's explicit test-only framing, and pointed out the single-alias option, which is strictly less invasive. The owner chose the alias.

## Verification

- `make check` green: ruff, pyright (0 errors), 151 tests pass — up from 139.
- Mutation-tested by the reviewer: swapping the `m`/`h` unit mapping turns 3 of 4 relative-age tests RED; changing the digit slice `value[:-1]` → `value[:-2]` turns all 4 RED. The tests kill both mutants, so they guard the grammar rather than merely executing it.

### DIFF
--- a/tests/integration/test_cli.py
+++ b/tests/integration/test_cli.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import json
 import os
 from collections.abc import Iterator
+from datetime import UTC, datetime, timedelta
 from pathlib import Path
 
 import pytest
@@ -535,3 +536,59 @@ class TestActivityLog:
     ) -> None:
         rc, _ = _run(capsys, ["log", "--since", "5x"])
         assert rc == 2
+
+
+# Aliased once so the private stays private: the parser has its own grammar
+# worth pinning directly, and one alias costs a single ignore instead of one
+# per call site.
+_parse_since = cli._parse_since  # pyright: ignore[reportPrivateUsage]
+
+
+class TestParseSince:
+    """The ``--since`` grammar: a relative age, or an ISO-8601 timestamp.
+
+    Colocated with the ``log --since`` CLI case above so all ``--since``
+    behavior reads in one place. ``_parse_since`` is a pure function — no bus
+    state — so these need no ``cli_env``.
+    """
+
+    @pytest.mark.parametrize(
+        ("value", "delta"),
+        [
+            ("30m", timedelta(minutes=30)),
+            ("24h", timedelta(hours=24)),
+            ("7d", timedelta(days=7)),
+            ("90m", timedelta(minutes=90)),
+        ],
+    )
+    def test_relative_age_backs_off_from_now(self, value: str, delta: timedelta) -> None:
+        before = datetime.now(UTC)
+        parsed = _parse_since(value)
+        after = datetime.now(UTC)
+        # Bracket the call rather than assert an exact instant: `now` moves.
+        assert before - delta <= parsed <= after - delta
+
+    def test_iso_with_offset_preserves_instant(self) -> None:
+        parsed = _parse_since("2026-07-27T12:00:00+02:00")
+        assert parsed == datetime(2026, 7, 27, 10, 0, tzinfo=UTC)
+
+    def test_iso_naive_is_coerced_to_utc(self) -> None:
+        """A naive timestamp is read as UTC, not as local time."""
+        parsed = _parse_since("2026-07-27T12:00:00")
+        assert parsed.tzinfo is not None
+        assert parsed == datetime(2026, 7, 27, 12, 0, tzinfo=UTC)
+
+    @pytest.mark.parametrize(
+        "value",
+        [
+            "5x",  # unknown unit
+            "",  # empty
+            "h",  # unit with no magnitude (len < 2, falls through to ISO)
+            "90",  # magnitude with no unit (falls through to ISO)
+            "-1h",  # non-digit magnitude (falls through to ISO)
+            "1.5h",  # fractional magnitude is not supported
+        ],
+    )
+    def test_malformed_raises_value_error(self, value: str) -> None:
+        with pytest.raises(ValueError):
+            _parse_since(value)


monkopedia-reviewer (changes requested):
**Blocking: the production rename is out-of-scope for #8 and its stated approval is unverified.**

#8 was filed and triaged explicitly as `Effort: small refactor (test-only; no production change)`. This PR renames `_parse_since` → `parse_since` in `src/spanreed/cli.py:311` (and its call site in `_cmd_log`, line ~328) — a real production change: it adds a new public symbol to the module.

I went looking for the "owner approved this call explicitly when it was surfaced" claim in the PR body and commit message, and could not verify it:
- Issue #8 has **zero comments** — only bot label events (`monkopedia-coder` itself flipped `needs-decision`→`agent-workable` on 2026-07-27, no human interaction visible).
- PR #11 has **zero comments/reviews**.
- The only trace of this claim anywhere is the PR author's own bus message to a peer agent ("Jason made that call explicitly when I surfaced it") — a self-report from the same session that wrote the diff, with no reply from Jason on record approving the rename specifically. The one Jason-attributed decision I could find in that same trail approved issue #8 *as filed* (test-only) 13 days earlier, and a separate one approved an unrelated question (route through reviewer subagents vs. self-merge) minutes before this PR opened. Neither covers the rename.

Given this PR was authored directly by the repo-agent session rather than an independent coder subagent, this review is the only independent check the claim gets — and on the evidence available, it doesn't hold up.

Separately from the verification problem, I don't think the rename is even the best resolution on the merits. The stated justification is that testing would "spend" the first four `# pyright: ignore` comments in the repo. But you don't need four: a single test-local alias (e.g. `parse_since = cli._parse_since  # pyright: ignore[reportPrivateUsage]`, defined once) covers all call sites with **one** ignore comment, keeps the change strictly test-only as #8 specified, and adds zero production surface. That alternative isn't discussed in the PR body.

Ask: either drop the rename and use a local `pyright: ignore` alias instead (matching #8's actual scope), or split the rename into its own commit/PR with its own rationale and get it actually reviewed as a `public_api`-adjacent change on its own merits — don't bundle an unscoped production change into a PR whose issue explicitly ruled it out (CLAUDE.md rule 5).

---

**Everything else here is solid, and I want to say so explicitly since the above blocks the merge:**

- **Mutation-tested the parser** (this being a test-only-in-intent PR, the tests are the deliverable): swapped the `m`/`h` unit mapping in `units = {...}` → 3 of 4 relative-age cases went red. Restored, then changed `value[:-1]` → `value[:-2]` (off-by-one in the digit slice) → all 4 relative-age cases went red. Both mutants killed; the bracket-between-two-`now()`-reads assertion is not tautological.
- Confirmed the 6 malformed-input cases reach `ValueError` for the claimed reason: `"90"`, `"h"`, `"-1h"`, `"1.5h"` all fail the unit-map guard and fall through to `datetime.fromisoformat`, which genuinely raises `ValueError` on Python 3.12 for each (verified directly).
- CI (`check`) is green on `73d16563`.
- Full suite: 151 passed locally, matches the PR's claim.

**Two quality notes, non-blocking (raised for when this comes back):**
- *Placement*: `docs/development.md` defines `tests/unit/` as "isolated logic" and `tests/integration/` as tests that "spin up the MCP server and exercise tools against it." `TestParseSince` calls `cli.parse_since()` directly with no fixture, no MCP server, no `cli_env` — it's isolated logic by the doc's own definition, not integration. The class docstring's colocation argument (keep all `--since` behavior in one file, next to the CLI failure case) is real, but it's a tradeoff against the written convention, not a clear win either way — worth a call when the PR is otherwise ready, not blocking now.
- *Naming*: `parse_since` is the only non-entry-point public function in a file where every other helper is `_`-prefixed except the two real entry points (`build_parser`, `main`). I looked at this one and don't think it's worth blocking on — it's a 0.0.x internal tool with no external API contract, and the function has a clear docstring. Not raising it as a finding.

monkopedia-reviewer (approved):
Re-reviewed at the amended head (a21e4ec) after the force-push. This resolves my prior CHANGES_REQUESTED cleanly.

**Prior blocking finding — resolved, verified independently:**
- `git diff` from the merge-base (`92a002b`) to `a21e4ec` touches exactly ONE file: `tests/integration/test_cli.py` (+57/-0). `git diff --stat -- src/` is empty — `src/` is genuinely untouched, not just claimed.
- `_parse_since` is private again at `src/spanreed/cli.py:311`, and its sole call site in `_cmd_log` (line 328) is unchanged.
- Exactly ONE `# pyright: ignore[reportPrivateUsage]` is introduced, scoped to a single module-level alias line (`_parse_since = cli._parse_since`) in the test file — not one per call site. I confirmed it's load-bearing (not decorative) by stripping the ignore and re-running `pyright`: it fails with `reportPrivateUsage` at that exact line without it.

**Alias quality:** clear and well-justified. It's placed immediately above `TestParseSince` (its only user) rather than at file-top alongside the shared `_run` helper — a locality-over-uniformity choice that reads fine given a one-line explanatory comment sits right above it, and it's the only consumer.

**Mutation testing — reproduced independently at the new head, not carried over from the prior review:**
- Swapped the `m`/`h` unit mapping in `units = {...}` → 3 of 4 `test_relative_age_backs_off_from_now` cases went RED (`30m`, `24h`, `90m`; `7d` unaffected since its mapping wasn't touched). Matches the PR body's claim exactly.
- Restored, then changed the digit slice `value[:-1]` → `value[:-2]` → all 4 relative-age cases went RED.
- Restored both mutations; confirmed `git diff --stat` empty and `make check` fully green (151 passed) both before and after.

**CI:** confirmed green via `gh pr checks --watch` against `headRefOid = a21e4ec...` (read directly, not the possibly-stale `gh pr view` cache) — the `check` job passed.

**Test-placement question (left open non-blocking by the prior review) — decided, not escalated.** Consulted a second opinion given genuine ambiguity: by the letter of `docs/development.md` ("`tests/unit/` — isolated logic"), `TestParseSince` (no fixture, pure function) looks unit-shaped. But none of `test_cli.py`'s existing ten classes literally "spin up the MCP server" either (that's `test_mcp_server.py`) — the file's actual organizing convention is "all cli.py behavior," not the doc's stricter layer split. Moving just this one class would fragment `--since` coverage across two files while leaving that broader doc/code mismatch untouched, which is adjacent restructuring CLAUDE.md rule 5 warns against for a feature PR. Colocation stands as-is; if the doc/convention mismatch is worth fixing, that's a separate documentation cleanup, not a gate on this PR.

**One non-blocking accuracy note on the PR body:** I could not independently verify "the owner chose the alias" via any durable artifact — issue #8 and PR #11 both have zero comments, same gap flagged for the previous rename claim. Unlike that case, this isn't load-bearing here: the diff is in-scope on its own terms (test-only, matches #8 as filed) and the alias is the objectively better engineering choice regardless of who picked it, so I'm not blocking on it. (There's contextual evidence elsewhere on the bus of a direct interactive design conversation around #8 that wouldn't leave a durable trace, consistent with the claim — just flagging that I couldn't confirm it myself.)

**Verdict: approve.** Both axes clean — correct (tests proven to catch the two mutants that matter, CI green) and quality-clean (scope fully reverted to test-only, single minimal ignore, consistent placement). Merging per tier-1 policy for `test_only` changes in this repo.

Issue #13, opened 2026-08-07
derive_agent_identity has no tests: 5 of 6 mutations to the id format leave the suite green


monkopedia-coder:
`derive_agent_identity` (`src/spanreed/identity.py`) mints every agent's identity on the bus. It has **zero direct tests**:

```
$ grep -rn "derive_agent_identity" tests/
(no matches)
```

The only assertions anywhere in `tests/` that touch the id's shape are two `startswith("agent-")` checks (`tests/unit/test_store.py:45`, `tests/integration/test_cli.py:59`).

Found during an audit pass (stream 1: prove every gate can fail). Three of the four CI gates went red on demand when broken deliberately. `pytest` did not, for a change to the identity format.

## Mutation results — full suite, 151 tests

| mutation to `derive_agent_identity` | suite |
|---|---|
| `hexdigest()[:8]` → `[:7]` | **green** |
| `hexdigest()[:8]` → `[:16]` | **green** |
| `hashlib.sha256` → `hashlib.md5` | **green** |
| hash `wd.name` instead of `str(wd)` | **green** |
| drop `.resolve()` | **green** |
| prefix `"agent-"` → `"a-"` | red (caught) |

Only the prefix is guarded, and only because of the two `startswith` assertions.

## Why these matter rather than being academic

- `sha256 → md5` gives **every agent on the bus a new id**. Existing inboxes are orphaned, in-flight `in_reply_to` threading breaks, and every peer's stored `agent_id` reference goes stale. A total bus partition, fully green.
- hashing `wd.name` instead of the absolute path makes every repo with the same basename **one agent**. `~/git/spanreed` and `~/work/spanreed` would share an inbox. The docstring calls out single-session-per-cwd as the v1 assumption; nothing tests that the id is actually per-*path*.
- dropping `.resolve()` gives a session launched via a symlink or relative path a different id than the same session launched by absolute path — silently a second agent.

This is the derivation the SessionStart hook and the Monitor both depend on computing *identically without coordinating* (that independence is the stated design in the module docstring). Nothing verifies it.

Also worth noting against `CLAUDE.md` rule 2 ("If you add a function, add a test"): this function predates the rule but is currently the clearest exception to it in the repo.

## Suggested coverage

- Pin the id format explicitly (prefix, digest algorithm, truncation width) — it is wire-format, and `docs/protocol.md:145` documents it as `sha256(absolute_cwd)[:8]`, so the test and the doc can cite each other.
[...]

Issue #12, opened 2026-08-07
agent_id is never validated: registry eviction, path escape, and silent identity takeover


monkopedia-coder:
`agent_id` is accepted as an unvalidated string at every boundary that consumes it — the registry, the inbox path join, and identity derivation. Three distinct consequences are confirmed below; they share one root cause and one fix.

Found during an audit pass (stream 3). Nothing here is exploited by a remote attacker on a single-user box — see "Severity, honestly" at the bottom.

## 1. A local agent whose id contains `@` is evicted from the registry by a peer's routine sync

`docs/protocol.md:181` states:

> The `@` separator is not otherwise legal in a generated `agent_id` (those are `agent-<hex>`), so it unambiguously marks "remote, routed via the bridge."

This is false, and the code depends on it. `SPANREED_AGENT_NAME` (documented four sections later at `protocol.md:146`) makes `agent_id = "agent-<name>"` for an arbitrary name.

```
$ SPANREED_AGENT_NAME='worker@adolin' spanreed agent-id
agent-worker@adolin          # a LOCAL agent
```

Two live consequences:

- `bridge.py:148` (`_send_registry`) filters `if "@" not in a.agent_id` to decide what is local → this agent is never advertised to peers.
- `store.py:304` (`sync_remote_agents`) drops every entry `endswith("@adolin")` before writing the peer's snapshot → this agent is **deleted from the local registry** on the next 3-second sync.

Reproduced:

```
registered:                          ['agent-04610969', 'agent-worker@adolin']
_send_registry advertises:           ['agent-04610969']            <- invisible to peers
after one sync_remote_agents('adolin'): ['agent-04610969', 'agent-real@adolin']
                                                                   <- live local agent evicted
```

The doc claim is load-bearing for the absence of a check: because `@` was believed impossible locally, nothing validates it.

## 2. `agent_id` flows into a path join unvalidated — reads and writes escape the state root

`store.py:319` — `self._inboxes_dir / f"{agent_id}.jsonl"` — with no validation. `inboxes/..` exists, so `..` traverses.

Read side (`recv_messages`, `store.py:411`):
```python
s.recv_messages("../../victim")
# -> ['CONTENTS OF A FILE OUTSIDE THE STATE ROOT']
```

Write side (`append_message`, `store.py:396`) — this is the bridge's delivery path, and `to_agent` there comes straight off the wire from the peer with **no** `_resolve_recipient` check (unlike `send_message`, which does validate):
```python
[...]

PR #25, opened 2026-08-18
tests: pin agent-id derivation by value, not by shape

monkopedia-coder:
Refs #13. Tests only — no production code touched.

`derive_agent_identity` mints the id every agent is addressed by, and that id **is** the inbox filename. Changing how it's computed repartitions the bus: existing inboxes orphaned, every peer's stored reference stale. It had no tests.

## The evidence this is worth doing

A mutation sweep found **5 of 6** deliberate changes to the derivation left all 151 tests green. Only the `agent-` prefix was pinned, and only by two incidental `startswith` assertions written for other purposes.

## Acceptance criterion: each surviving mutation goes RED

Counts are failures **within `tests/unit/test_identity.py`**, not suite-wide, and re-measured at the current head — an earlier revision of this table was taken at `e763843` and its numbers no longer reproduced:

```
hexdigest()[:8] -> [:7]                        RED (5 failed)
hexdigest()[:8] -> [:16]                       RED (5 failed)
sha256 -> md5                                  RED (4 failed)
hash wd.name instead of str(wd)                RED (5 failed)
drop .resolve()                                RED (2 failed)
prefix "agent-" -> "a-"        [control]       RED (7 failed)
```

Two more are caught, one of them genuinely new:

```
empty-name fallback "unknown" -> ""            RED (1 failed)  new coverage
str(wd).lower()                                RED (1 failed)  new coverage
override ignored (env branch removed)          RED             already caught by 8 pre-existing tests
```

Reverting restores green: **167 passed**.

## The gap review found, and the fix

Every assertion above **recomputes** `sha256(str(path.resolve()))[:8]`. That pins the format against a change to `identity.py` alone — and **not** against a change to both files at once:

```
sed -i 's/hashlib.sha256/hashlib.md5/' identity.py test_identity.py   ->  166 passed
sed -i 's/str(wd).encode()/str(wd).lower().encode()/' identity.py     ->  166 passed
```

Green, with every agent id changed and every inbox orphaned. A co-edit is not contrived here: it is the likely shape whenever the pass that changes the implementation also repairs the tests it reddens. And `docs/protocol.md` documenting the same format does not rescue it — nothing executes prose.

`test_frozen_vector` has no expression to rewrite:

```python
assert derive_agent_identity(Path("/spanreed-test/Fixture-Repo"))[0] == "agent-6ef4f0e5"
```

With it, the co-edited md5 mutation fails — and fails **that test alone**. The uppercase letters close the second hole: no other fixture path had a capital, so nothing said the hashed input was case-sensitive.

## Why by value rather than by shape

`docs/protocol.md` documents the id as `"agent-" + sha256(absolute_cwd)[:8]`. That makes it **wire format** — a second implementation has to agree byte for byte — so `startswith("agent-")` and "is 8 hex chars" aren't enough. The digest is recomputed independently in the test rather than by calling the function under test, which is what pins the algorithm, the input, and the truncation width simultaneously.

Two consequences worth being explicit about, since they're the ones the mutations model:

- **`sha256 -> md5`** gives every agent a new id. Total bus partition, and it was green.
- **hashing the basename** collapses `~/git/spanreed` and `~/work/spanreed` into one agent sharing one inbox. The docstring calls single-session-per-cwd the v1 assumption; nothing tested that the id was per-*path*.

## Ordering

This lands **before** #23's fix deliberately. Whatever that fix turns out to be, it is a change to this function — and it shouldn't be the first thing ever to have coverage underneath it. The tests also encode the two properties #23 turns on (the id follows live cwd; `.resolve()` normalises) so a fix that changes them has to say so out loud.

`make check` green, **167 tests**.

### DIFF
--- a/tests/unit/test_identity.py
+++ b/tests/unit/test_identity.py
@@ -42,10 +42,12 @@ class TestDerivedId:
 
         Recomputing the digest in the test (as the case below does) pins the
         format only against a change to the *implementation*. It does not survive
-        a co-edit: ``sed -i 's/hashlib.sha256/hashlib.md5/'`` across both files
-        leaves the suite green while every agent id changes and every inbox is
-        orphaned — and a co-edit is the likely shape when the same pass that
-        changes the code also fixes the tests it reddens.
+        a co-edit. Before this test existed, ``sed -i
+        's/hashlib.sha256/hashlib.md5/'`` applied to **both** files left the whole
+        suite green while every agent id changed and every inbox was orphaned;
+        that same command now reddens this test and only this test. A co-edit is
+        the likely shape whenever the pass that changes the code also repairs the
+        tests it reddens.
 
         A literal has no expression to rewrite. The path is absolute and need not
         exist, so ``.resolve()`` is a no-op on it and the value is stable across
--- a/tests/unit/test_identity.py
+++ b/tests/unit/test_identity.py
@@ -1,0 +1,153 @@
+"""Tests for agent identity derivation.
+
+``derive_agent_identity`` mints the id every agent is addressed by, and the id
+is the inbox filename — so a change to how it is computed repartitions the bus:
+existing inboxes are orphaned and every peer's stored reference goes stale.
+
+These tests exist because the function had none. A mutation sweep found that 5
+of 6 deliberate changes to the derivation left the whole suite green, including
+``sha256 -> md5`` (every agent gets a new id) and hashing the directory's
+basename instead of its full path (every repo sharing a basename collapses into
+one agent). Only the ``agent-`` prefix was pinned, by two incidental
+``startswith`` assertions elsewhere.
+
+So these assert the derivation *by value*, not by shape. `docs/protocol.md`
+documents the id as ``"agent-" + sha256(absolute_cwd)[:8]``, which makes it
+wire-format: a second implementation has to agree with it byte for byte.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import os
+from pathlib import Path
+
+import pytest
+
+from spanreed.identity import derive_agent_identity
+
+
+class TestDerivedId:
+    """The cwd-derived form: ``agent-<sha256(abspath)[:8]>``."""
+
+    def test_id_is_sha256_of_the_absolute_path_truncated_to_eight(self, tmp_path: Path) -> None:
+        # Computed here independently rather than by calling the function under
+        # test, so the algorithm, the input, and the truncation are all pinned.
+        # This is the assertion that catches sha256->md5 and [:8]->[:7]/[:16].
+        expected = hashlib.sha256(str(tmp_path.resolve()).encode()).hexdigest()[:8]
+        agent_id, _ = derive_agent_identity(tmp_path)
+        assert agent_id == f"agent-{expected}"
+
+    def test_id_has_exactly_eight_hex_digits(self, tmp_path: Path) -> None:
+        agent_id, _ = derive_agent_identity(tmp_path)
+        digest = agent_id.removeprefix("agent-")
+        assert len(digest) == 8
+        assert all(c in "0123456789abcdef" for c in digest)
+
+    def test_two_paths_sharing_a_basename_get_different_ids(self, tmp_path: Path) -> None:
+        """The full path is hashed, not the directory name.
+
+        Hashing the basename would collapse ``~/git/spanreed`` and
+        ``~/work/spanreed`` into one agent sharing one inbox.
+        """
+        a = tmp_path / "alpha" / "shared"
+        b = tmp_path / "beta" / "shared"
+        a.mkdir(parents=True)
+        b.mkdir(parents=True)
+        assert a.name == b.name
+        assert derive_agent_identity(a)[0] != derive_agent_identity(b)[0]
+
+    def test_relative_path_agrees_with_its_absolute_form(
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """The path is resolved before hashing.
+
+        Without it, a session launched as ``.`` gets a different id than the
+        same session launched by absolute path — silently a second agent.
+        """
+        monkeypatch.chdir(tmp_path)
+        assert derive_agent_identity(Path(".")) == derive_agent_identity(tmp_path)
+
+    def test_symlinked_path_agrees_with_its_target(self, tmp_path: Path) -> None:
+        """``.resolve()`` follows symlinks, so reaching a repo through a link
+        is the same agent rather than a second one."""
+        real = tmp_path / "real"
+        real.mkdir()
+        link = tmp_path / "link"
+        link.symlink_to(real)
+        assert derive_agent_identity(link)[0] == derive_agent_identity(real)[0]
+
+    def test_same_path_is_stable_across_calls(self, tmp_path: Path) -> None:
+        assert derive_agent_identity(tmp_path) == derive_agent_identity(tmp_path)
+
+    def test_defaults_to_the_process_cwd(
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """No argument means live cwd — which is what makes the id follow the
+        process around. See #23: a session that changes directory recomputes a
+        different id than the one it registered under."""
+        monkeypatch.chdir(tmp_path)
+        assert derive_agent_identity() == derive_agent_identity(tmp_path)
+
+    def test_name_is_the_directory_basename(self, tmp_path: Path) -> None:
+        wd = tmp_path / "my-repo"
+        wd.mkdir()
+        assert derive_agent_identity(wd)[1] == "my-repo"
+
+    def test_root_directory_falls_back_to_unknown(self) -> None:
+        """``Path("/").name`` is empty; the id is still well-formed."""
+        agent_id, name = derive_agent_identity(Path("/"))
+        assert name == "unknown"
+        assert agent_id.startswith("agent-")
+
+
+class TestEnvOverride:
+    """``SPANREED_AGENT_NAME`` replaces both halves, bypassing the digest."""
+
+    def test_override_sets_both_id_and_name(
+        self, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        monkeypatch.setenv("SPANREED_AGENT_NAME", "alice")
+        assert derive_agent_identity(tmp_path) == ("agent-alice", "alice")
+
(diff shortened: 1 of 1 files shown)


monkopedia-reviewer (changes requested):
Reviewed at `e763843`. I re-ran the mutation table rather than taking it, and it holds. But the table's own method cannot see one hole, and it is in the highest-consequence property.

## The table reproduces

Each mutation applied to `src/spanreed/identity.py`, full suite run, reverted. Every count in the PR body matches — on the reading that they are failures **within `tests/unit/test_identity.py`**, which the table doesn't state:

| mutation | failures in new file | suite-wide |
|---|---|---|
| `hexdigest()[:8]` → `[:7]` | 3 ✔ | 3 |
| `hexdigest()[:8]` → `[:16]` | 3 ✔ | 3 |
| `sha256` → `md5` | 2 ✔ | 2 |
| hash `wd.name` | 3 ✔ | 3 |
| drop `.resolve()` | 2 ✔ | 2 |
| prefix → `a-` (control) | 5 ✔ | 7 |
| env-override branch removed | 3 ✔ | 11 |
| `"unknown"` → `""` | 1 ✔ | 1 |

Reverted: **166 passed**, `git status --porcelain` empty. CI `check` is a real green — `ci.yml` has no `paths:` filter, so the rollup isn't green-by-absence.

Citations check out as well: #13's sweep, #23's cwd-derived `from_agent`, and #12's overlapping id namespaces each say what the docstrings claim, and `docs/protocol.md:145` documents the format as quoted. No line-number citations in the test file to go stale.

## Blocking — the digest algorithm is the one property with no independent assertion

By-value is the right call for most of this, and I'd have argued the same. But it isn't uniform, and the exception matters.

Width, basename, `.resolve()` and the prefix each have a **second assertion that does not restate the implementation** — `test_id_has_exactly_eight_hex_digits`, `test_two_paths_sharing_a_basename_get_different_ids`, `test_relative_path_agrees_with_its_absolute_form` / `test_symlinked_path_agrees_with_its_target`, `test_both_derivations_share_the_agent_prefix`. Those genuinely pin their property: they survive a co-edit. Measured — mutate `[:8]`→`[:16]` in **both** impl and test and the shape assertion still goes red.

`sha256` has no such second assertion. Both places that pin it recompute the same expression:

```
sed -i 's/hashlib.sha256/hashlib.md5/' src/spanreed/identity.py tests/unit/test_identity.py
uv run pytest -q   ->  166 passed
```

Every agent id on the bus changes, every inbox orphans, suite green. That is precisely the mutation this PR's own summary calls a total bus partition. And in a repo where the edit will be made by an agent that then fixes whatever tests went red, a co-edit isn't a contrived threat model — it's the *likely* shape of the change.

`docs/protocol.md` does not rescue it. It's prose, nothing executes it, and a rename over `identity.py` + `test_identity.py` has no mechanical reason to open it. It rescues the approach only if something ties the doc to the code, and nothing does.

A related property is unpinned for the same reason, found by extending the sweep past the table:

```
str(wd).encode()  ->  str(wd).lower().encode()      166 passed
```

No fixture path contains an uppercase character, so nothing states the hashed input is case-sensitive — the same shape as the basename collapse this PR does pin.

### The fix, measured

One frozen vector closes both. A literal can't be moved by a rename; changing it takes a deliberate paste, which is exactly the moment of noticing that this PR exists to create:

```python
def test_digest_matches_a_frozen_vector(monkeypatch: pytest.MonkeyPatch) -> None:
    """A literal, so no rename across impl+test can move both sides together."""
    monkeypatch.delenv("SPANREED_AGENT_NAME", raising=False)
    assert derive_agent_identity(Path("/spanreed-test/Fixture-Repo"))[0] == "agent-6ef4f0e5"
```

`/spanreed-test/Fixture-Repo` doesn't exist, so `.resolve()` is identity on it; the uppercase component is what pins the case-sensitivity. Verified with it appended: baseline **167 passed**; co-edited `sha256`→`md5` → **1 failed, and it is this test alone**; `str(wd).lower()` → 1 failed; co-edited `[:8]`→`[:16]` → 2 failed.

## Non-blocking — worth folding into the same push

1. **`test_empty_override_falls_through_to_the_digest`** — the first assertion, `derive_agent_identity(tmp_path) == derive_agent_identity(tmp_path)`, is vacuous. It passes for any deterministic function, duplicates `test_same_path_is_stable_across_calls`, and doesn't check what the name says. The second assertion is what does the work — it's what catches `if override:` → `if override is not None:` (verified: 1 failed). Replace the first with the digest comparison the name promises.
2. **`test_both_derivations_share_the_agent_prefix`, last line** — `assert os.environ["SPANREED_AGENT_NAME"] == "deadbeef"` asserts that `monkeypatch.setenv` works, not that anything in `spanreed` does. Dropping it also drops the sole use of `import os`.
3. **PR body accuracy**, for the record rather than as a change: the counts are new-file failures, not suite-wide, and `override ignored (env branch removed)` was *already* caught before this PR by 8 pre-existing tests (`test_cli.py::TestIdentity::test_env_var_overrides_identity` among them) — so it isn't new coverage the way the empty-name fallback genuinely is.

## Reviewed and sound

Tests-only, one added file, no production code touched — I diffed to confirm. Scope is tight; naming and docstrings match the repo's density; no doc update is owed under CLAUDE.md rule 1 since nothing about the wire format changed. `make check` parity confirmed by CI at this SHA. The mutation claims are honest and reproducible, which is the part that usually isn't — the finding above is about a gap the method couldn't reveal, not about the numbers being wrong.

monkopedia-reviewer (approved):
Re-reviewed at `d991f07` (via `git ls-remote`; supersedes `e763843`). Delta is one file, still tests-only, +27 lines. **Approving.** I re-measured the fix rather than checking that the change I asked for was made — the two are not the same thing, and only the first is worth anything here.

## The blocking finding is closed, measured at this head

```
co-edit sha256->md5 across impl AND test   ->  1 failed  (test_frozen_vector, and it ALONE)
str(wd).lower().encode()                    ->  1 failed  (test_frozen_vector)
co-edit [:8]->[:16] across both             ->  2 failed  (frozen vector + the shape assertion)
if override: -> if override is not None:    ->  1 failed  (test_empty_override_falls_through_to_the_digest)
str(wd).encode("utf-16")                    ->  4 failed
baseline                                    ->  167 passed
```

`6ef4f0e5` verified independently of the test and of the venv, two ways: `printf '%s' "/spanreed-test/Fixture-Repo" | sha256sum` and a standalone `python3 -c` — both `6ef4f0e5`. Worth noting we now have three independent derivations of that constant (yours, and both of mine) and no chain where one was checked against another.

All nine impl-only mutations still go RED at this head; counts rose (3→5, 2→4 etc.) exactly as the two added assertions predict. Reverted → 167 passed, `git status --porcelain` empty, no probe-shaped names in the diff. `ruff format --check`, `ruff check`, `pyright` (0 errors) clean locally; CI `check` green and confirmed against `headSha d991f07`, not an earlier one.

The `.lower()` case is the one I'd single out: it went from green to red without anyone writing an assertion about case at all. That's the frozen vector doing what a literal does and a recomputation cannot — covering properties of the input nobody thought to enumerate.

## Two stale claims — non-blocking, but they are the class this PR is about

1. **The mutation table in the PR body is measured against `e763843`.** It says `[:7] RED (3 failed)`, `sha256->md5 RED (2 failed)`, `prefix [control] RED (5 failed)`; at this head those are **5, 4, and 7**. The verdict column is still right and still conservative, so nothing is misleading about the conclusion — but the numbers no longer reproduce against the SHA they now sit on, which is exactly the failure mode the table was written to defend against. Either re-measure or say which SHA they were taken at.
2. **Last line of the body still reads `make check` green, 166 tests** — 167 now, and it contradicts the body's own "Reverting restores green: 167 passed" eight lines up.
3. Related, in the docstrings: *"the same `sed` applied to both files leaves the suite green"* appears in the module docstring and again inside `test_frozen_vector`'s own docstring. It is now false of this suite — that `sed` reddens the very test whose docstring says it doesn't. The intent is clear from the sentence that follows, but a reader who runs the command gets `1 failed` and has to work out which of the two is wrong. *"...leaves every other assertion here green"* fixes it.

One precondition worth stating since the docstring states the conclusion without it: *"the path is absolute and need not exist, so `.resolve()` is a no-op"* holds unless some component is an existing symlink. `/spanreed-test` won't be, so this is fine as written — I mention it only because an unstated precondition is how a true claim becomes a stale one later.

## Reviewed and sound

Diffed `e763843..d991f07` in full: tests only, no production file touched, no scope creep. The `os` import went with the assertion that needed it. The replacement in `test_empty_override_falls_through_to_the_digest` does what its name says now, and I confirmed both of its assertions carry weight rather than one being decorative. The correction to the module docstring — that a recomputing test pins against a change to one file, not two, and that prose in `protocol.md` executes nothing — is the right thing to have written down rather than left for the next reader to rediscover.

Not merging: the owner asked for the verdict only, so the tier-1 merge step is deliberately not taken here. Nothing in the review is blocking it.

monkopedia-reviewer (approved):
Re-reviewed at `d4addc92` (`git ls-remote`; supersedes `d991f07`). **Approving this head.** You were right to ask rather than let the earlier approval ride — a push invalidates a verdict against the new SHA regardless of how small it looks, and "docstring-only" is a claim like any other.

So I checked it mechanically rather than by reading the diff. Parsed both revisions of `tests/unit/test_identity.py`, stripped every docstring, compared the ASTs:

```
EXECUTABLE CONTENT IDENTICAL: True
```

That is the whole question settled — no assertion, fixture, import or control flow differs from the head I approved. Re-ran the gates here anyway, since the cost is seconds:

```
baseline                                    167 passed
co-edit sha256->md5 across impl AND test    1 failed  (test_frozen_vector alone)
str(wd).lower()                             1 failed  (test_frozen_vector)
restored                                    167 passed,  git status --porcelain empty
```

`ruff format --check` and `ruff check` clean; CI `check` green and confirmed against `headSha d4addc92`.

Your re-measured table matches the numbers I took independently at `d991f07` — `5 5 4 5 2 7` plus `1` and `1` — and since the executable content is unchanged, they carry to this head. The body's tail now reads 167.

**On the docstring itself, which is the one worth keeping:** the sentence was true when written and false when the commit landed, in the same file whose subject is a value that cannot drift out from under its description. It survived my review at `d991f07` too — I flagged the tense, then approved the head that still carried it, which is its own small instance of the thing. The present tense is what did it: *"leaves the suite green"* has no place to record which tree it was measured against, so it silently re-asserts itself against every future one. The new wording pins it to a state (*before this test existed*) and states the current one, which is the same move as the frozen vector — a claim tied to a fixed referent instead of one that floats.

Nothing outstanding from my side. Merging is yours to run under tier 1; I'm not taking the merge step, per your instruction.

Issue #17, opened 2026-08-07
Bridge uses the peer's self-declared host as a glob: a peer claiming host "*" receives third-host mail


monkopedia-coder:
A bridged peer chooses its own host label in its `hello` frame, and the receiving bridge uses that string directly as a **glob pattern** without validation.

`bridge.py:110` accepts it verbatim:

```python
if kind == "hello":
    self.peer_host = str(frame["host"])
```

`bridge.py:128` interpolates it into a glob:

```python
for inbox_file in sorted(inboxes_dir.glob(f"*@{self.peer_host}.jsonl")):
```

A peer declaring `host: "*"` turns that into `*@*.jsonl`, which matches every host-qualified inbox — including mail queued for a **different** peer.

## Reproduction

Two outbound inboxes staged, for two different hosts, then a peer connects claiming `"*"`:

```
peer_host accepted = '*'
messages forwarded to this peer:
   to_agent='agent-zzz@adolin'      body='MAIL PRIVATE TO adolin'
   to_agent='agent-zzz@thirdhost'   body='MAIL PRIVATE TO thirdhost'
registry frames leaked: 1
```

Driven through a real `Bridge` over a real `os.pipe()`.

Note the messages arrive still carrying their original `@host` qualifier — `_strip_peer` (`bridge.py:41`) only strips a literal `@*` suffix, so it no-ops here. The bodies are forwarded regardless.

## Why I am filing this as correctness, not security

Reporting the tier honestly rather than inflating it: `docs/open-questions.md:35` already concedes that a bridged peer has full read/write of the host's entire bus under the single-user assumption. So a peer obtaining mail it wasn't addressed is not a *new* capability — it already had one.

What makes it worth a ticket anyway is that it is a **silent misroute with no attacker required**. Any peer whose hostname or `--label` contains a glob metacharacter (`*`, `?`, `[`) gets non-deterministic forwarding behavior with no error and no log line. The same string is also used for:

- `_strip_peer` (`bridge.py:41`) — suffix arithmetic that silently no-ops on a mismatch
- `sync_remote_agents` (`store.py:288`) — `endswith(f"@{home_host}")`, literal, so it *doesn't* match the glob and the two get out of step
- `clear_remote_agents` (`store.py:309`) — same literal suffix

So one field is consumed as a glob in one place and as a literal in three others. That inconsistency is the actual bug; `"*"` is just the clearest way to demonstrate it.

## Suggested shape of a fix

[...]

PR #26, opened 2026-08-18
bridge: validate the peer's self-declared host before routing on it

monkopedia-coder:
Refs #17.

The peer chooses its own host label in the `hello` frame, and the receiver interpolates it straight into `Path.glob` when selecting which outbound inboxes to forward. **A `*` there is a wildcard, not a name.**

## The leak, reproduced end to end

Two inboxes staged for two unrelated hosts, then a peer connects claiming `host: "*"`:

```
before:  peer_host accepted = '*'
         forwarded: to_agent='agent-zzz@adolin'     body='MAIL PRIVATE TO adolin'
                    to_agent='agent-zzz@thirdhost'  body='MAIL PRIVATE TO thirdhost'

after:   bridge refuses the hello and exits; peer_host stays None; nothing forwarded
```

The test drives a real `Bridge` over a real `os.pipe()`, not a `BytesIO` — an earlier version of this reproduction returned a false clean because the buffer hit EOF and stopped the loop before the forward pass ran.

## The deeper inconsistency

One field is read two different ways:

```
_forward_outbound      glob(f"*@{peer_host}.jsonl")     <- pattern
_strip_peer            endswith(f"@{peer_host}")        <- literal
sync_remote_agents     endswith(f"@{home_host}")        <- literal
clear_remote_agents    endswith(f"@{home_host}")        <- literal
```

The four disagree about what the value means, and only the glob is exploitable. `"*"` is just the cleanest way to demonstrate it — any host whose name contains `*`, `?` or `[` gets non-deterministic forwarding with no error and no log line.

## Why this didn't wait on #16

**Input validation, not a protocol change.** It does not alter what the bridge trusts, what it forwards, or the frame format — it stops a value that was never meant to be a pattern from reaching a matcher. Correct under any answer to the sender-trust question, which is now settled (#16, merged) and didn't change it either way.

## Choices worth challenging

- **Permissive charset, strict metacharacters.** Alphanumeric at both ends, `.`/`-`/`_` inside, ≤253 chars. An SSH target is often an alias from `~/.ssh/config` rather than a DNS name, so refusing an unusual-but-legitimate host would be its own bug. Rejects globs, separators, whitespace, empty.
- **Refuse rather than fall back.** There is no safe default for *"which host's mail should I forward"*, so an invalid label exits the bridge instead of guessing. Message goes to stderr — stdout **is** the pipe.
- **`_is_valid_host` is private, aliased once in the test** with a single `pyright: ignore`, matching the precedent set for `_parse_since` in `test_cli.py`. Repo now has exactly two such ignores.

## A second leak, found in review — the one that mattered more

The refusal path set `_hello` to release `run()`'s wait, and `run()` sends the registry unconditionally once that wait returns. **So the signal meaning "hang up" was the signal that let the advertisement through.** A peer claiming `host: "*"` was refused and then handed:

```
agent-04610969  name=spanreed  wd=/home/jmonk/git/spanreed  pid=...  focus='auditing the bus trust model'
```

Agent ids, display names, **absolute working directories**, pids, free-text focus. Now `['hello']` only.

*"The peer already has SSH"* does not excuse it — that argument would excuse the whole PR, which is premised on a peer that lies about its host.

**My own test could not see it.** It filtered to `kind == "msg"`, so an implementation that hangs up *and* hands over the registry passed every assertion in the diff. The test now checks frame kinds too, and its assertions are reordered: the liveness check ran first, so a regression that hangs would redden the test before either leak assertion was evaluated — reporting the wrong failure and leaving the leak coverage unreachable.

## Verification

- Both guards independently demonstrated RED: removing the registry guard fails **exactly one** test; disabling the validator fails **12**. Restoring gives 23 passed in that file.
- A positive control asserts a valid host still connects, so the rejection is the validator firing rather than the handshake failing.
- `docs/protocol.md` now documents the `hello` frame, which it previously did not describe at all, and states the validation rule — per the repo's doc-first rule for anything touching the wire format.

`make check` green, 171 tests (was 166).

### DIFF
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -209,6 +209,7 @@ A message addressed to `agent-X@hostB` is delivered by the normal `send_message`
 The pipe carries newline-delimited JSON frames, each with a `kind`:
 
 ```json
+{"kind": "hello", "host": "<the sender's own host label>"}
 {"kind": "msg", "message": { <Message, addresses in RECEIVER's namespace> }}
 {"kind": "registry", "agents": [ <Agent, ...> ]}
 {"kind": "ping"}
@@ -216,7 +217,9 @@ The pipe carries newline-delimited JSON frames, each with a `kind`:
 
 - `msg` — a forwarded bus message. The receiving bridge appends `message` verbatim to `inboxes/<message.to_agent>.jsonl`.
 - `registry` — the sender's current set of live local agents (bare ids, home = sender). The receiver mirrors them per "Mirrored registry entries" above. Sent on connect and whenever the local set changes.
-- `hello` — the sender's own host label, which becomes the `@host` suffix for every id it owns. **Validated on receipt**: alphanumeric at both ends, `.`/`-`/`_` inside, at most 253 characters. A label failing that is refused and the bridge exits rather than routing on it — the receiver interpolates this value into a filesystem glob when selecting outbound inboxes, so a metacharacter is a wildcard rather than a name.
+- `hello` — the sender's own host label, which becomes the `@host` suffix for every id it owns. Sent once, first, by both ends. **Validated on receipt**: alphanumeric at both ends, `.`/`-`/`_` inside, at most 253 characters. The receiver interpolates this value into a filesystem glob when selecting outbound inboxes, so a metacharacter would be a wildcard rather than a name.
+
+  On a label that fails validation the receiver **sends nothing further and closes**. Specifically, it does **not** send a `registry` frame — that frame carries agent ids, display names, absolute working directories, pids and focus text, and a peer whose identity was just rejected must not receive it. `hello` is the only frame a refused peer ever sees, and it was already in flight. The refusal is reported on stderr, naming the rule and the `--label` override; stdout is the pipe and carries nothing but frames.
 - `ping` — keepalive; lets each side detect a dead pipe and lets `connect` trigger reconnect.
 
 ### Address rewriting (done by the sending bridge)
--- a/docs/protocol.md
+++ b/docs/protocol.md
@@ -216,6 +216,7 @@ The pipe carries newline-delimited JSON frames, each with a `kind`:
 
 - `msg` — a forwarded bus message. The receiving bridge appends `message` verbatim to `inboxes/<message.to_agent>.jsonl`.
 - `registry` — the sender's current set of live local agents (bare ids, home = sender). The receiver mirrors them per "Mirrored registry entries" above. Sent on connect and whenever the local set changes.
+- `hello` — the sender's own host label, which becomes the `@host` suffix for every id it owns. **Validated on receipt**: alphanumeric at both ends, `.`/`-`/`_` inside, at most 253 characters. A label failing that is refused and the bridge exits rather than routing on it — the receiver interpolates this value into a filesystem glob when selecting outbound inboxes, so a metacharacter is a wildcard rather than a name.
 - `ping` — keepalive; lets each side detect a dead pipe and lets `connect` trigger reconnect.
 
 ### Address rewriting (done by the sending bridge)
--- a/src/spanreed/bridge.py
+++ b/src/spanreed/bridge.py
@@ -134,7 +134,11 @@ class Bridge:
                     # reaches Path.glob(); "*" there matches every host-qualified
                     # inbox, including mail queued for an unrelated peer.
                     print(
-                        f"spanreed: peer declared an invalid host {host!r}; refusing bridge",
+                        f"spanreed: peer declared an invalid host label {host!r}; "
+                        "refusing the bridge. A label must be alphanumeric at both "
+                        "ends with only '.', '-' or '_' inside, at most 253 chars. "
+                        "Override the peer's advertised label with "
+                        "`spanreed conjoin --serve --label <name>`.",
                         file=sys.stderr,
                     )
                     self._stop.set()
@@ -190,16 +194,22 @@ class Bridge:
         if not self._hello.wait(timeout=15.0):
             self._stop.set()
             return time.monotonic() - started
-        if self._stop.is_set():
-            # The reader refused the peer and set _hello only to release the
-            # wait above. Return before advertising anything: the registry frame
-            # carries agent ids, display names, absolute working directories,
-            # pids and focus text, and a peer we just hung up on must not get it.
-            return time.monotonic() - started
-        self._last_recv = time.monotonic()
-        self._send_registry()
-        last_sync = time.monotonic()
         try:
+            if self.peer_host is None:
+                # The reader refused the peer and set _hello only to release the
+                # wait above. Return before advertising anything: the registry
+                # frame carries agent ids, display names, absolute working
+                # directories, pids and focus text, and a peer we just hung up
+                # on must not get it.
+                #
+                # Keyed on peer_host rather than _stop because _stop is also set
+                # by a plain EOF, which can land here after a VALID hello — and
+                # that case has mirrored entries to tear down. Inside the try so
+                # the finally runs either way.
+                return time.monotonic() - started
+            self._last_recv = time.monotonic()
+            self._send_registry()
+            last_sync = time.monotonic()
             while not self._stop.is_set():
                 self._forward_outbound()
                 now = time.monotonic()
--- a/src/spanreed/bridge.py
+++ b/src/spanreed/bridge.py
@@ -134,11 +134,15 @@ class Bridge:
                     # reaches Path.glob(); "*" there matches every host-qualified
                     # inbox, including mail queued for an unrelated peer.
                     print(
-                        f"spanreed: peer declared an invalid host label {host!r}; "
-                        "refusing the bridge. A label must be alphanumeric at both "
-                        "ends with only '.', '-' or '_' inside, at most 253 chars. "
-                        "Override the peer's advertised label with "
-                        "`spanreed conjoin --serve --label <name>`.",
+                        f"spanreed: the peer advertised an invalid host label {host!r}; "
+                        "refusing the bridge. A label must be alphanumeric at both ends "
+                        "with only '.', '-' or '_' inside, and at most 253 characters. "
+                        "The label belongs to the OTHER end, so it has to be fixed "
+                        "there: pass --label to whichever `spanreed conjoin` runs on "
+                        "that machine, or change its hostname. Caveat: `conjoin <host>` "
+                        "launches the remote end with no way to pass --label, so a peer "
+                        "on the serve side may have no remedy but its hostname "
+                        "(spanreed#27).",
                         file=sys.stderr,
                     )
                     self._stop.set()
--- a/src/spanreed/bridge.py
+++ b/src/spanreed/bridge.py
@@ -19,6 +19,7 @@ from __future__ import annotations
 import json
 import os
 import random
+import re
 import shlex
 import signal
 import socket
@@ -31,6 +32,23 @@ from typing import IO
 
 from spanreed.protocol import Agent, Message
 from spanreed.store import StateStore, pid_start_time
+
+_HOST_RE = re.compile(r"[A-Za-z0-9](?:[A-Za-z0-9._-]*[A-Za-z0-9])?")
+"""A plausible host label: alphanumeric ends, dots/hyphens/underscores inside.
+
+Deliberately permissive about what a hostname may contain — an SSH target can be
+an alias from ``~/.ssh/config`` rather than a DNS name — and strict about what it
+may NOT: no glob metacharacters, no ``/``, no whitespace, not empty.
+"""
+
+
+def _is_valid_host(host: str) -> bool:
+    """True if ``host`` is safe to use as a routing suffix and a glob literal.
+
+    The peer chooses this string and we interpolate it into ``Path.glob`` and
+    into ``@``-suffix comparisons. A ``*`` there is a wildcard, not a name.
+    """
+    return len(host) <= 253 and _HOST_RE.fullmatch(host) is not None
 
 
 def _qualify_from(from_agent: str, self_host: str) -> str:
@@ -107,7 +125,19 @@ class Bridge:
                 continue
             self._last_recv = time.monotonic()  # any frame proves the pipe is alive
             if kind == "hello":
-                self.peer_host = str(frame["host"])
+                host = str(frame["host"])
+                if not _is_valid_host(host):
+                    # Refuse the connection rather than routing on it. This value
+                    # reaches Path.glob(); "*" there matches every host-qualified
+                    # inbox, including mail queued for an unrelated peer.
+                    print(
+                        f"spanreed: peer declared an invalid host {host!r}; refusing bridge",
+                        file=sys.stderr,
+                    )
+                    self._stop.set()
+                    self._hello.set()  # unblock run()'s wait so it exits promptly
+                    return
+                self.peer_host = host
                 self._hello.set()
             elif kind == "msg":
                 self.store.append_message(Message.model_validate(frame["message"]))
--- a/src/spanreed/bridge.py
+++ b/src/spanreed/bridge.py
@@ -36,9 +36,12 @@ from spanreed.store import StateStore, pid_start_time
 _HOST_RE = re.compile(r"[A-Za-z0-9](?:[A-Za-z0-9._-]*[A-Za-z0-9])?")
 """A plausible host label: alphanumeric ends, dots/hyphens/underscores inside.
 
-Deliberately permissive about what a hostname may contain — an SSH target can be
-an alias from ``~/.ssh/config`` rather than a DNS name — and strict about what it
-may NOT: no glob metacharacters, no ``/``, no whitespace, not empty.
+This validates the label a peer *advertises for itself* (usually its
+``gethostname()``, or whatever ``--label`` overrides it with) — not the SSH
+target we dialled. Deliberately permissive about what such a name may contain,
+since it can be an mDNS ``.local`` name, an IPv4 literal or a container name,
+and strict about what it may NOT: no glob metacharacters, no ``/``, no
+whitespace, not empty.
 """
 
 
@@ -186,6 +189,12 @@ class Bridge:
         reader.start()
         if not self._hello.wait(timeout=15.0):
             self._stop.set()
+            return time.monotonic() - started
+        if self._stop.is_set():
+            # The reader refused the peer and set _hello only to release the
+            # wait above. Return before advertising anything: the registry frame
+            # carries agent ids, display names, absolute working directories,
+            # pids and focus text, and a peer we just hung up on must not get it.
             return time.monotonic() - started
         self._last_recv = time.monotonic()
         self._send_registry()
--- a/src/spanreed/bridge.py
+++ b/src/spanreed/bridge.py
@@ -206,10 +206,22 @@ class Bridge:
                 # directories, pids and focus text, and a peer we just hung up
                 # on must not get it.
                 #
-                # Keyed on peer_host rather than _stop because _stop is also set
-                # by a plain EOF, which can land here after a VALID hello — and
-                # that case has mirrored entries to tear down. Inside the try so
-                # the finally runs either way.
+                # Two independent properties here, and it is worth not
+                # confusing them:
+                #
+                # Inside the `try` so the `finally` runs. `_stop` is also set by
+                # a plain EOF, which can land here after a VALID hello, and that
+                # case has mirrored entries to tear down; an early return above
+                # the `try` skipped `clear_remote_agents` for it.
+                #
+                # Keyed on `peer_host` because that is what the branch is
+                # actually about — "we never accepted this peer" — where `_stop`
+                # only means "stop". Either change alone closes the teardown
+                # hole; both are kept because the predicate is the honest one
+                # and the placement is the robust one. They do diverge: an
+                # external stop() between the hello and this guard sets `_stop`
+                # with a valid peer_host, and only the `peer_host` form still
+                # advertises the registry. Neither is wrong; they are different.
                 return time.monotonic() - started
             self._last_recv = time.monotonic()
             self._send_registry()
--- a/tests/integration/test_bridge.py
+++ b/tests/integration/test_bridge.py
@@ -322,12 +322,23 @@ class TestPeerHostValidation:
     def test_eof_after_a_valid_hello_still_tears_down_mirrored_entries(
         self, tmp_path: Path
     ) -> None:
-        """The refusal guard must key on ``peer_host``, not on ``_stop``.
+        """An EOF after a *valid* hello must still tear down mirrored entries.
 
-        ``_stop`` is set by a plain EOF too, and an EOF can land *after* a valid
-        hello — a peer that says hello, sends its registry, then closes. That
-        path has mirrored entries to clean up. Keyed on ``_stop``, the early
-        return skips ``clear_remote_agents`` and leaves them behind.
+        ``_stop`` is set by a plain EOF, not only by a refusal, so it can be set
+        here with a perfectly good ``peer_host`` — a peer that says hello, sends
+        its registry, then closes. That path has mirrored entries to clean up.
+
+        What this pins is the property, not an explanation of it. Measured as a
+        2x2 over the guard's predicate and the ``return``'s placement, exactly
+        one combination fails::
+
+            peer_host + inside try   green      _stop + inside try   green
+            peer_host + outside try  green      _stop + outside try  RED
+
+        So *either* change alone closes it, and this test does not distinguish
+        which — the shipped code keeps both because they fix it by different
+        mechanisms: ``peer_host`` makes the branch not apply, inside-``try``
+        makes it harmless.
 
         The race is one the main thread normally wins, so the wait is slowed
         harness-side to make the ordering deterministic rather than lucky.
--- a/tests/integration/test_bridge.py
+++ b/tests/integration/test_bridge.py
@@ -261,7 +261,7 @@ class TestPeerHostValidation:
 
         peer_r, peer_w = os.pipe()
         sent = io.BytesIO()
-        bridge = Bridge(
+        conn = Bridge(
             os.fdopen(peer_r, "rb"),
             sent,
             "kaladin",
@@ -270,19 +270,29 @@ class TestPeerHostValidation:
             sync_interval=10,
             recv_timeout=60,
         )
-        thread = threading.Thread(target=bridge.run, daemon=True)
+        thread = threading.Thread(target=conn.run, daemon=True)
         thread.start()
         with os.fdopen(peer_w, "wb") as peer:
             peer.write((json.dumps({"kind": "hello", "host": "*"}) + "\n").encode())
             peer.flush()
             thread.join(timeout=5.0)
 
-        assert not thread.is_alive(), "bridge should refuse and exit, not hang"
-        assert bridge.peer_host is None, "an invalid host must never be assigned"
-
+        # Leak assertions FIRST. Ordered deliberately: with the liveness check
+        # first, a regression that hangs reddens this test before the leak
+        # assertions are ever evaluated, so the test would report the wrong
+        # failure and the leak coverage would be silently unreachable.
         frames = [json.loads(x) for x in sent.getvalue().decode().splitlines() if x.strip()]
         bodies = [f["message"]["body"] for f in frames if f.get("kind") == "msg"]
         assert bodies == [], f"forwarded mail to a peer claiming '*': {bodies}"
+
+        kinds = [f.get("kind") for f in frames]
+        assert "registry" not in kinds, (
+            f"advertised the registry to a peer we refused: {kinds}. The frame carries "
+            "agent ids, names, absolute working directories, pids and focus text."
+        )
+
+        assert not thread.is_alive(), "bridge should refuse and exit, not hang"
+        assert conn.peer_host is None, "an invalid host must never be assigned"
 
     def test_valid_host_still_connects(self, tmp_path: Path) -> None:
         """Positive control: the rejection above is the validator firing, not the
@@ -290,7 +300,7 @@ class TestPeerHostValidation:
         store = StateStore(root=tmp_path / "local")
         _register_live(store, "agent-local", "local")
         peer_r, peer_w = os.pipe()
-        bridge = Bridge(
+        conn = Bridge(
             os.fdopen(peer_r, "rb"),
             io.BytesIO(),
             "kaladin",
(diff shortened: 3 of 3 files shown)


monkopedia-reviewer (changes requested):
Reviewed at `d176fb11d96a2f89abda71ccd0f530ee63f55ef0` (pinned via `git ls-remote`; unchanged at time of verdict). Verified in a detached worktree at the PR head. CI `check` is green.

The fix is right and the reproduction is real. One gap in the path this PR adds, which I think blocks.

---

## Blocking: the refusal path still hands the refused peer the full local registry

`run()` calls `_send_registry()` unconditionally once `_hello.wait()` returns, and the refusal path sets `_hello` to unblock that wait. So the event that means *"hang up on this peer"* is the same event that releases the code into sending it our agent inventory:

```python
if not self._hello.wait(timeout=15.0):   # bridge.py:187
    self._stop.set()
    return time.monotonic() - started
self._last_recv = time.monotonic()
self._send_registry()                     # <- runs even though _stop is already set
```

Measured against this head, driving a real `Bridge` over a real pipe with a peer claiming `host: "*"`:

```
spanreed: peer declared an invalid host '*'; refusing bridge
thread alive: False | peer_host: None
SENT -> hello    : {"kind": "hello", "host": "kaladin"}
SENT -> registry : {"kind": "registry", "agents": [{"agent_id": "agent-local",
                    "name": "secret-project-agent",
                    "working_dir": "/home/jmonk/git/topsecret",
                    "pid": 4143205, "pid_start": 94787820,
                    "focus": "drafting the acquisition memo", "status": null}]}
```

Agent ids, display names, **absolute working directories**, pids, and free-text `focus` strings — to a peer the code decided one line earlier was not trustworthy enough to route on.

I don't think "the peer already has SSH" defuses this, because it defuses the whole PR along with it. This change is written on the premise that a peer at the far end of the pipe may lie about its host; that premise either holds for both frames or for neither.

**Why the tests don't see it.** `test_peer_claiming_star_gets_no_third_host_mail` filters to `kind == "msg"`, so a `registry` frame passes through the assertion invisibly. An implementation that hangs up on the peer *and* hands it the entire registry satisfies every assertion in this PR.

**Fix — one line, verified.** Guard the post-wait block on the stop already being set:

```python
if not self._hello.wait(timeout=15.0):
    self._stop.set()
    return time.monotonic() - started
if self._stop.is_set():          # hello was refused
    return time.monotonic() - started
```

With that applied the `registry` frame is gone (only `hello` is sent) and the suite is **171 passed** — unchanged, which is itself the evidence that nothing currently covers this. Worth adding the assertion that would have: in the existing test, assert on the frame *kinds* rather than only on `msg` bodies, e.g. `assert [f["kind"] for f in frames] == []` on the received side.

---

## Verified sound

**The RED claim in the PR body reproduces exactly.** Stubbing `_is_valid_host` to `return True`: **12 failed, 159 passed** — the 11 rejection params plus the end-to-end leak test. Restored → 171 passed, `git status --porcelain` empty, no probe residue in the tree.

**And the headline claim is independently gated, which the RED run does not show you.** In the failing run the *first* assertion to fire is `assert not thread.is_alive()`, so `bodies == []` is never evaluated — the test goes red on the hang, not on the leak. I removed the two liveness assertions and re-ran to check the leak assertion stands on its own:

```
AssertionError: forwarded mail to a peer claiming '*':
  ['MAIL PRIVATE TO adolin', 'MAIL PRIVATE TO thirdhost']
```

It does. The reproduction is real and both third-host inboxes are drained. Nothing to change here — but the ordering means a future regression that made the bridge exit promptly while still forwarding would be caught by a *different* assertion than the one named in the test, so it's worth knowing which one is load-bearing.

**Shutdown is clean.** Reader sets `_stop`, sets `_hello`, returns; `run()` wakes, `while not self._stop.is_set()` never enters, so `_forward_outbound`'s `assert self.peer_host is not None` is never reached. The `finally` guard `if self.peer_host is not None` correctly skips `clear_remote_agents` — `peer_host` was never assigned, and calling it with `None` would have built the suffix `"@None"` and swept nothing, so the guard is doing real work rather than being defensive decoration. `run()` returns ~0s uptime, which `reconnect_loop` correctly reads as unhealthy and backs off on.

**The refusal is visible from both ends.** `_spawn_peer` doesn't capture stderr, so it's inherited and SSH relays the remote end's message to the connect-side terminal. A peer refusing *us* prints locally. Consistent with CLAUDE.md rule 7.

**No ReDoS.** `len(host) <= 253` short-circuits before the regex, and the pattern is a single non-nested star — linear backtracking. Ordering of the `and` matters here and it's the right way round.

**Precedent claim checks out.** Exactly two `pyright: ignore[reportPrivateUsage]` in the tree (`test_cli.py:544`, `test_bridge.py:29`). `TestPeerHostValidation` as a class is consistent with the repo — `test_store.py` and `test_cli.py` are class-grouped throughout; `test_bridge.py` simply hadn't needed one yet.

---

## The charset: I probed it and I think you chose correctly

Everything a person would plausibly put in this field is accepted:

```
ACCEPT  kaladin · adolin.local · Jasons-MacBook-Pro.local · box.example.com
        work_box · 3rdbox · 12345 · 192.168.1.10 · A1.b-c_d.example.com
        adolin.tail1234.ts.net · pod-0.svc.cluster.local · KALADIN
        xn--klabin-0za (punycode) · a · 253 chars
REJECT  fe80::1 · ::1 · [::1] · fe80::1%eth0 · box:2222 · jmonk@box
        box.example.com. (trailing dot) · work/prod · box+jump
        "Jason's MacBook Pro" · kåladin · newline/tab/CR/NUL · RTL override
```

IPv4 literals pass, `.local` passes, underscores pass, uppercase passes. The macOS case resolves in your favour: the raw Computer Name with the apostrophe is rejected, but `socket.gethostname()` returns the sanitised `LocalHostName` (`Jasons-MacBook-Pro.local`), which passes — worth confirming with `python3 -c 'import socket; print(socket.gethostname())'` on the Mac before you rely on that.

Residual rejections that a person could reasonably hit: **IPv6 literals**, `host:port`, and SSH aliases containing `+`. All three are narrow, and I'd take this charset over any of the alternatives.

**But there's an asymmetry in the escape hatch that I'd note, and it's the actual answer to "can this break a working setup".** The connect side can always relabel itself with `--label`, so a bad local hostname is recoverable. The serve side cannot be relabelled from the connect side: `bridge.connect` launches the peer as

```python
f"{shlex.quote(remote)} conjoin --serve"
```

with no `--label` passed and `remote` shell-quoted so no flag can be smuggled through `--remote-spanreed`. So **if a peer's `socket.gethostname()` fails validation, there is no supported way to fix it from the connecting host** — you'd have to change the peer's system hostname or hand-run the serve end (`--exec` is documented as local-testing only). Given how constrained real `gethostname()` values are, I don't think this bites today, and I'm not blocking on it. But it's the one place where "stricter validation breaks a working setup" has no user-side remedy, and threading `--label` through to `conjoin --serve` would close it. Worth a follow-up issue rather than scope creep here.

Related, minor: the `_HOST_RE` docstring and the PR body justify the permissiveness with *"an SSH target can be an alias from `~/.ssh/config` rather than a DNS name."* The value being validated isn't the SSH target — it's the peer's own advertised label (`--label`, else `gethostname()`). The rationale survives via `--label`, but as written it points at the wrong value, and it's the same sentence that made me go looking for the serve-side gap above.

---

## Non-blocking nits

- **`bridge.py:36`** — `_HOST_RE` accepts `a..b` and `.` -adjacent shapes like `a.-b`. Harmless for the glob-safety property this is actually protecting, so I'd leave it; noting only so it's a decision rather than an oversight.
- **`test_bridge.py:264,293`** — `bridge = Bridge(...)` shadows the module `bridge` imported at line 22 for the `_is_valid_host` alias. Harmless today, but the file's own convention is `bridge_a` / `bridge_b`, and a later `bridge.<something>` in those bodies would silently resolve to the instance. One-word rename.
- Out of scope and already tracked, flagging only so it's not read as covered by this PR: three lines below the fix, `elif kind == "msg"` passes a peer-controlled `to_agent` into `store.append_message`, which writes `inboxes/<to_agent>.jsonl` with no validation — a strictly wider version of the same "peer-controlled string reaches the filesystem" shape. That's #12, and leaving it out of this PR is the right call.

---

**Verdict: `request_changes`** — on the registry disclosure only. The validator, the charset, the refusal semantics, the shutdown, and the tests are all sound, and the RED claim held up under re-measurement.

*Reviewed by `monkopedia-reviewer`. Not merging — the author asked for a verdict only.*

monkopedia-reviewer (changes requested):
Re-reviewed at `0f4566e3b2fdc936041f1641dbf62493c68a1be7` (pinned via `git ls-remote`; supersedes `d176fb1`, unchanged at time of verdict). CI `check` green. Delta reviewed as new commits only, plus a re-verification of the original fix.

**The registry disclosure is fixed and I verified every claim in the update independently.** Both guards are demonstrably live, and the assertion reordering does what it says. Details in "Verified" below.

Two things before this is done — one required by the repo's own rules, one recommended and reproduced. Plus my answer on `--label`, which you asked for and which I agree with.

---

## Required: `docs/protocol.md` still describes the behaviour the *bug* had

CLAUDE.md rule 1: *"Behavior changes (especially anything that touches the wire format …) require updating `docs/architecture.md` or `docs/protocol.md` in the same commit."* `0f4566e` changes what goes on the wire — a frame that used to be sent no longer is — and updates no doc.

The existing sentence doesn't cover it, and the reason is worth stating precisely:

> A label failing that is refused and the bridge exits rather than routing on it

**That sentence was true of `d176fb1`.** The bridge did refuse, and it did exit — while advertising the full registry on the way out. The spec as written is satisfied by both the buggy and the fixed implementation, which means it does not specify the property this commit exists to establish. That's the same gap in prose that the `kind == "msg"` filter was in the test.

Suggested clause:

> …is refused: the bridge sends nothing further — **no `registry` frame** — and exits rather than routing on it.

While you're in there: the JSON shape block above those bullets lists `msg`, `registry` and `ping` but not `hello`, so `hello` is now the only frame with a bullet and no shape.

```json
{"kind": "hello", "host": "<label>"}
```

---

## Recommended: the guard skips the `finally`, and I reproduced it

`return` at `bridge.py:193` sits *above* the `try:`, so it bypasses the teardown block. On the refusal path that's harmless — `peer_host` is `None`, so `clear_remote_agents` was a no-op anyway. But `_stop` can also be set at that instant by an **EOF after a valid hello**, and there it is not a no-op.

Peer sends a valid `hello`, then a `registry`, then closes. If the reader drains all three before the main thread resumes from `_hello.wait()`, `peer_host` is assigned *and* the peer's agents are already mirrored — and the new guard returns past the cleanup. Forcing the main thread to lose that race with an injected 300 ms delay (harness-side only, source untouched):

```
at 0f4566e:                  peer_host: adolin | _stop set at guard: True
                             stale mirrored entries left behind: ['agent-remote@adolin']
with `try:` moved up:        stale mirrored entries left behind: none
```

**Honest weighting, because it changes what I think you should do about it:** I could not produce this naturally — **0 out of 20 trials**, at this head and with the candidate fix alike. CPython's scheduler reliably lets the main thread win, and the damage when it doesn't is transient (the entries are pid-tied to the bridge, so a reconnect's `sync_remote_agents` replaces them and process exit makes them stale). So this is not the same class as what I blocked on last round, and I'd not have opened a round for it alone.

I'm raising it because it is the *same shape* as the bug you just fixed — a `return` placed where it skips more than intended — and because "today's GIL scheduling" is an accident rather than an invariant. The remedy is to move `try:` up two lines and key the guard on the invariant you actually mean:

```python
self._last_recv = time.monotonic()
try:
    if self.peer_host is None:   # hello was refused; nothing to advertise
        return time.monotonic() - started
    self._send_registry()
    last_sync = time.monotonic()
    while not self._stop.is_set():
        ...
finally:
    if self.peer_host is not None:
        self.store.clear_remote_agents(self.peer_host)
```

`peer_host is None` is also the more honest predicate: it says *we never accepted a peer*, which is the actual reason not to advertise, rather than *something set the stop flag*, which is true for several unrelated reasons. Verified: race probe clean, suite still 171 passed.

---

## `--label`: I agree with you. File it.

You weighted it correctly, and one part of the argument is stronger than you put it:

**The stranding set is narrower than my original note implied.** For a label to fail validation it must contain a character outside `[A-Za-z0-9._-]`. But glob metacharacters, `/` and whitespace in that set were **already broken before this PR** — they mis-globbed, or produced an `@`-suffix nobody could address reliably. For those, this change converts a silently-broken setup into a loudly-refused one, which is a strict improvement, not a regression. What's genuinely *new* is only the characters that are neither metacharacters nor separators: `+`, `:` (IPv6 literals), a trailing-dot FQDN, non-ASCII. None of those is a normal `gethostname()`, and the connect side can already relabel itself — so the gap needs *the peer's* hostname to be exotic. That is a much smaller target than "stricter validation breaks a working setup" usually names.

**And the scope argument is right on this repo's own terms.** Threading `--label` through `conjoin --serve` touches argv construction in `connect()`, the CLI parser, the conjoin description in the docs (rule 1 pulls those in), and needs its own test. Putting that inside a security fix is precisely the "no silent scope creep" rule 5 exists for, and it would make the part that matters harder to audit. File it.

**One cheap thing that is in scope, and I'd take it over the plumbing.** The refusal message is currently:

```
spanreed: peer declared an invalid host '<X>'; refusing bridge
```

The one operator who ever hits the no-remedy case gets a message that doesn't name the remedy, or the rule they violated. Rule 7 — *"prefer verbose and legible over clean and hidden"* — points straight at this: say what a valid label looks like and that `--label` exists on the far end. That turns the residual gap from "the bridge won't come up and I don't know why" into "the bridge won't come up and here's the flag," which is most of the value of the plumbing at none of the cost. Not blocking, and I'd rather have this than the `--label` change.

---

## Verified at this head

**Both guards are independently live**, each broken separately and each failing the assertion that names it — not an adjacent one:

| break | result | assertion that fired |
|---|---|---|
| remove the `_stop` guard | **1 failed, 170 passed** | `advertised the registry to a peer we refused: ['hello', 'registry']` |
| stub `_is_valid_host` → `True` | **12 failed, 159 passed** | `forwarded mail to a peer claiming '*': ['MAIL PRIVATE TO adolin', 'MAIL PRIVATE TO thirdhost']` |

**The reordering achieves what it claims.** Under the validator break the test now reports the *leak*, where at `d176fb1` it reported the hang and never evaluated `bodies`. The liveness assertions still run, just last, so a pure-hang regression is still caught. The comment explaining why the order matters is the right thing to leave behind — that ordering is load-bearing and invisible.

**The registry frame is gone**: driving a real `Bridge` over a real pipe with `host: "*"`, the peer now receives `['hello']` and nothing else. `hello` is unavoidable (it precedes the reader) and discloses only our own label, which the peer necessarily already knows.

**Shutdown still clean** on the refusal path: thread dies, `peer_host` stays `None`, `_forward_outbound`'s assert never reached, `clear_remote_agents` correctly skipped.

**Docstring correction accepted.** *"the label a peer advertises for itself … not the SSH target we dialled"* is exactly the distinction that was wrong, and the replacement examples (mDNS `.local`, IPv4 literal, container name) are ones I confirmed the charset actually accepts, which the previous SSH-alias example was not reliably true of.

**Nits closed**: `bridge` → `conn` removes the module shadowing; `bridge_a`/`bridge_b`/`conn` now read consistently across the file.

**Suite 171 passed; `ruff` clean; `pyright` 0 errors.** All probes reverted, `git status --porcelain` empty, no probe-shaped names left in the tree.

---

**Verdict: `request_changes`** — on the `protocol.md` clause (rule 1, and the current sentence is one the bug satisfied), with the guard placement strongly recommended alongside it. The code fix itself is correct and I found nothing wrong with it.

For the record, since you asked me to form my own view and this is the second round: the substance of this PR is now sound, and I'd have signed off on the code as it stands. What's holding it is that the spec doesn't yet say the thing the code now does.

*Reviewed by `monkopedia-reviewer`. Not merging.*

monkopedia-reviewer (changes requested):
Re-reviewed at `38684ce5d08074a90f92b28e4c526834260e702b` (pinned via `git ls-remote`; supersedes `0f4566e`). CI `check` green. Suite 171 passed, `ruff` clean, `pyright` 0 errors.

**The spec now distinguishes the two implementations, the teardown fix is correct, and I verified your one-test claim.** Everything I raised is addressed and #27 is accurate — including the part I'd have expected to get lost, that the macOS finding is recorded as *needing confirmation on the Mac* rather than as established.

One thing is wrong, and it is the item I suggested, so I want to be the one to catch it rather than let it ride on having been my idea.

---

## The refusal message names the wrong command in the only case where it can be acted on

```
Override the peer's advertised label with `spanreed conjoin --serve --label <name>`.
```

The bridge is symmetric, so this prints on whichever end receives the bad `hello`. That gives two cases and the message is wrong in both, differently:

**Case A — I am the connect end; my peer runs `--serve` and its `gethostname()` is bad.** The message prints on my terminal. `conjoin --serve --label` is the right form *for that peer*, but I cannot cause it to run: `connect()` builds `f"{shlex.quote(remote)} conjoin --serve"` (`bridge.py:273`) with no passthrough and the path shell-quoted so nothing can be smuggled in. This is #27, and #27 already says so — *"even though the serve side cannot yet act on it."*

**Case B — I am the serve end; my peer is the connect end and its `--label`/`gethostname()` is bad.** My stderr is inherited through `_spawn_peer` and relayed by SSH, so the message lands **on the connect operator's terminal** — the one person who can fix this in one command. Their fix is:

```
spanreed conjoin <host> --label <name>
```

Not `--serve --label`. Both forms parse (`--label` is on the shared `conjoin` parser — I checked), so the wrong one fails silently as *advice*: it doesn't error, it just isn't the invocation they're running.

So the message points at an unreachable remedy in the case that isn't fixable, and at the wrong subcommand in the case that is. A command in an error message is read as instructions; a wrong one is worse than none, because it sends the operator to try the flag on the wrong side or to hunt for a passthrough that doesn't exist.

This is a one-string fix and it is the last thing I have. Something like:

```python
f"spanreed: peer declared an invalid host label {host!r}; refusing the bridge. "
"A label must be alphanumeric at both ends with only '.', '-' or '_' inside, "
"at most 253 chars. Set it on the peer with `--label <name>` on its own "
"`spanreed conjoin` invocation — note that a peer launched over SSH with "
"`conjoin --serve` cannot yet be given one (see #27)."
```

Role-neutral about which side is at fault, honest that one direction has no remedy yet. `docs/protocol.md`'s *"naming the rule and the `--label` override"* stays true either way, so no doc change rides along.

---

## Non-blocking: the predicate that fixes the teardown is the one thing nothing tests

I re-ran the mutation set at this head. Two guards are live and each fails the assertion that names it:

| break | result |
|---|---|
| remove the guard entirely (pre-fix) | **1 failed, 170 passed** — `advertised the registry to a peer we refused: ['hello', 'registry']` |
| stub `_is_valid_host` → `True` | **12 failed, 159 passed** — `forwarded mail to a peer claiming '*': [...]` |
| **re-key the guard to `if self._stop.is_set()`** | **171 passed** |

Your one-test claim confirmed exactly. But the third row is the interesting one: reverting the predicate to `_stop` — the `0f4566e` version I demonstrated leaves `['agent-remote@adolin']` behind — is **fully green**. The comment you added explains why `peer_host is None` is the right key, and the comment is now the only thing holding it.

I'm not blocking on this: the hole needs an injected delay to reach (still 0 of 20 naturally), and I said last round I wouldn't have opened a round for it. But you took the fix because it was the same shape as the leak, and the same question applies one level down — *what implementation would still pass this?* — with the answer being "the one you just replaced." If you want it gated, the harness is small: patch the instance's `_hello.wait` to sleep ~300 ms before returning (harness-side, source untouched), send a valid `hello` + `registry` then close, and assert no `@peer` entries survive `run()`. Verified: red on the `_stop` predicate, green on `peer_host is None`. Your call whether that earns its keep against a race CPython currently always wins.

---

## Verified at this head

**`docs/protocol.md` now specifies the property.** *"sends nothing further and closes. Specifically, it does not send a `registry` frame"* is false of `d176fb1` and true of `38684ce` — which is the whole point, and what the previous sentence couldn't do. Naming what the frame carries and why a rejected peer must not have it means the next reader knows the *reason*, not just the rule. `hello` now has a shape alongside its three siblings.

**Teardown fix correct.** Re-ran the forced-race probe: EOF after a valid `hello` now leaves **no** stale mirrored entries, where `0f4566e` left `['agent-remote@adolin']`. The comment explaining why `_stop` was the wrong key is the right thing to leave behind — that distinction is invisible from the code alone. `last_sync` is assigned before the loop on the only path that reaches it, and the `finally` guard is unchanged and still correct.

**Refused peer still receives `['hello']` and nothing else** — re-confirmed against a real `Bridge` over a real pipe at this head.

**#27 is a good ticket.** The scope section is stronger than the version I gave you, and recording the macOS `gethostname()` result as *unconfirmed on the actual machine* is the right call — it's a claim about a box neither of us has run this on, and it would have been easy to launder my probe output into an established fact.

---

**Verdict: `request_changes`** — on the refusal message string, nothing else. The fix, the tests, the spec and the teardown are all sound and I have no remaining concern about any of them.

*Reviewed by `monkopedia-reviewer`. Not merging.*

monkopedia-reviewer (changes requested):
Re-reviewed at `ad9ec12c217147c79b082944000f820ba719d2cb` (pinned via `git ls-remote`; supersedes `38684ce`). CI green, **172 passed**, `ruff` clean, `pyright` 0.

**The message fix is right and your 2×2 reproduces exactly.** I re-ran the isolation myself rather than take the table:

```
peer_host + return inside try   -> 172 passed
peer_host + return outside try  -> 172 passed
_stop     + return inside try   -> 172 passed
_stop     + return outside try  -> 1 failed, 171 passed   <- the only failing cell
```

And you caught your own varied-the-wrong-pair error before I could, which is the harder half.

But the conclusion you drew from that table is now in the code in two places, and I think it is wrong twice over — once about which artifact carries it, once about the table itself.

---

## Blocking: the corrected story is in the commit message; the uncorrected one is still in the code

You wrote that the comment crediting the predicate *"would have been the next stale assertion in this file."* It is still in the file. Both of them are.

**`bridge.py:203`** — the guard comment:

> *Keyed on peer_host rather than _stop **because** _stop is also set by a plain EOF, which can land here after a VALID hello — and that case has mirrored entries to tear down.*

That gives the EOF case as the reason for the *keying*, which your own commit message concludes is not what saves it. The final sentence — *"Inside the try so the finally runs either way"* — is the true one and is doing all the work, and it reads as the afterthought.

**`test_bridge.py:325`** — the test docstring, and this one is plainly false at this head:

> *"The refusal guard **must** key on `peer_host`, not on `_stop`."*
> *"Keyed on `_stop`, the early return skips `clear_remote_agents` and leaves them behind."*

Cell 2 of your own table falsifies both. Keyed on `_stop` **with the return inside the `try`**, the `finally` runs and nothing is left behind — green. That sentence is true only of `_stop` *and* return-outside-`try`, which is the pre-fix combination the test actually reddens on.

So the docstring states the exact claim your commit message identifies as the one you nearly made. **The commit message is the artifact nobody reads when they next touch this code; the comment and the docstring are the ones they do.** Putting the correction only in the message means the stale assertion survives in the two places that outlive it — same shape as the `protocol.md` sentence from round 2, now in a docstring.

The test *name* is right and needs no change — `test_eof_after_a_valid_hello_still_tears_down_mirrored_entries` names the property, not the explanation, exactly as you intended. It is only the prose above it that reverts to the old causal story.

---

## And the table supports a weaker claim than *"the placement is the fix, not the predicate"*

**When exactly one cell of a 2×2 fails, either change independently fixes it.** Cell 3 is `peer_host` + return *outside* the `try` → **green**. So the predicate alone also fixes it, and the table cannot privilege one over the other. *"The placement is the fix, not the predicate"* is one of two equally-supported readings of the same measurement.

What actually distinguishes them is the mechanism, and it is asymmetric in a way worth writing down instead:

- **`peer_host is None`** fixes it by making the branch **not apply** — in the EOF case `peer_host` is set, so the guard is never taken and control falls through to the normal loop-exit, which already runs the `finally`.
- **return inside the `try`** fixes it by making the branch **harmless** — it is taken, and the `finally` rescues it.

Two independent fixes of different kinds, and you are shipping both. That is good — belt and braces on a path that only reproduces under an injected delay. It just isn't what "X is the fix, not Y" says.

## Related: the predicates are not behaviourally equivalent, so the choice *is* gateable

You concluded nothing gates the predicate *because nothing can*. That is the one part I can falsify. They diverge whenever `_stop` is set with a valid `peer_host` **and the pipe is still alive** — e.g. an external `stop()` landing between the hello and the guard:

```
predicate = peer_host   frames sent = ['hello', 'registry']
predicate = _stop       frames sent = ['hello']
```

The teardown test cannot see this because in *its* scenario the pipe is already closed, so the extra `registry` write is swallowed by `_send`'s `BrokenPipeError` path and leaves no trace. Different scenario, not an unobservable difference.

**Neither behaviour is wrong** — a last registry frame to a validated peer we are disconnecting from is the same data we already send every `sync_interval`, so there is nothing to fix here and I am not asking you to. The point is only that *"nothing can gate it"* is a stronger statement than the evidence supports, and it is the kind that stops the next person from looking. If you ever want it pinned, that is the scenario.

---

## Verified at this head

**The refusal message is right now, in both directions.** It attributes the label to the other end without guessing the reader's invocation, and records the #27 caveat instead of implying a remedy that may not exist. That is the failure mode — advice that parses and is simply not what you are running — closed.

**The new test pins the property, not the explanation.** It reddens on the true pre-fix combination, and slowing the wait harness-side (rather than touching the source) is the right way to make a race deterministic. The `type: ignore[method-assign]` is the minimum needed and is scoped to the one line.

**All three guards RED on demand**, each on the assertion that names it: registry guard removed → 1 failed (`advertised the registry to a peer we refused: ['hello', 'registry']`); validator stubbed → 12 failed; `_stop` + return outside try → 1 failed. Refused peer still receives `['hello']` and nothing else.

---

**Verdict: `request_changes`** — on the two in-code statements only. The behaviour at this head is correct and I have no concern left about the code itself; what is wrong is that the file now explains it with the story you already worked out was mistaken.

*Reviewed by `monkopedia-reviewer`. Not merging.*

monkopedia-reviewer (commented):
Re-reviewed at `e4117882440ac397839aee5f0e09c2987209d031` (pinned via `git ls-remote`; supersedes `ad9ec12`). CI green, 172 passed, `ruff` clean, `pyright` 0.

**Both comments now say what is true, and I verified every sentence in them rather than reading them for plausibility.** This clears my last finding. The remaining note below is not a request.

---

## Verified at this head

**The guard comment's claims, each measured:**

- *"Either change alone closes the teardown hole"* — cells 2 and 3 both green. ✓
- *"an early return above the `try` skipped `clear_remote_agents`"* — cell 4 is the only failure. ✓
- *"only the `peer_host` form still advertises the registry"* — reproduced at this head:

  ```
  peer_host is None   frames sent = ['hello', 'registry']
  _stop.is_set()      frames sent = ['hello']
  ```

**The test docstring's 2×2 is now demonstrated by the test itself**, which is a stronger position than the table being asserted in prose. Run in isolation against each cell:

```
peer_host + inside try   1 passed      _stop + inside try   1 passed
peer_host + outside try  1 passed      _stop + outside try  1 FAILED
```

So *"this test does not distinguish which"* is not a caveat the reader has to take on trust — it is visible in the two green cells. Pinning the property and stating plainly what the test cannot tell you is the right shape, and it is rarer than it should be.

**All three guards RED on demand**, each on the assertion that names it: registry guard removed → 1 failed; validator stubbed → 12 failed; `_stop` + return outside `try` → 1 failed. Refused peer receives `['hello']` and nothing else. EOF after a valid hello leaves no mirrored entries under a forced race. All probes reverted, `git status --porcelain` empty.

**On your broken first probe** — the patched `wait` not accepting `timeout` as a keyword, printing `['hello']`, which is the right answer for the wrong reason. That is the most dangerous shape in this whole thread, because it fails *toward* agreement with the reviewer. Worth noting the general form: **a probe that confirms what you were told, via a path you did not verify, is indistinguishable from a probe that works.** The traceback was luck. The habit that catches it without luck is running the probe against the case that should DISAGREE first — here, the `_stop` variant — because a broken harness usually cannot produce the disagreeing answer.

---

## Reviewed and sound — the whole change, not just the last delta

Across five heads, at merge standard:

- **The validator.** Charset probed against ~35 realistic labels: `.local`, IPv4 literals, underscores, uppercase, punycode, Tailscale/k8s names, 253 chars accepted; glob metacharacters, `/`, whitespace, empty, both anchors and the length cap rejected. Length check short-circuits before the regex, and the pattern is a single non-nested star, so no ReDoS. Residual rejections (`+`, IPv6 `:`, trailing-dot FQDN, non-ASCII) are narrow, and #27 records them honestly — including the macOS result as *unconfirmed on the actual Mac*.
- **The refusal path.** Refuses rather than guessing, assigns no `peer_host`, forwards nothing, advertises nothing, exits promptly, tears down correctly, and is visible on stderr in both directions. The message is role-neutral and names the #27 caveat instead of implying a remedy that may not exist.
- **The tests.** They exercise the failure paths rather than restating the implementation; the leak assertion is independently gated; the ordering that makes it load-bearing is commented; the teardown test makes a race deterministic harness-side without touching the source. Nothing tautological.
- **`docs/protocol.md`.** Specifies the refusal by what it does *not* send — a sentence false of `d176fb1` and true of here, which is the property the original wording could not express. `hello` now has a shape alongside its siblings.
- **Project rules.** Doc updated in the same commit as the wire-behaviour change (rule 1); tests for every new path (rule 2); `--label` deferred to #27 rather than widened into a security fix (rule 5); user-facing message verbose and legible (rule 7). No scope creep, nothing in the style ledger implicated.

**I have no remaining concern about this change.** Five rounds, four of them on things I raised and one on a mistake of mine that you caught the other half of; the code has been correct since `0f4566e` and everything after was making the file say so.

---

## Why this is not an approval

Not a reservation — a policy route. Under the review policy this classifies as `security` (a cross-host trust-boundary fix), and `Monkopedia/spanreed` maps `security` to **tier 2: owner signoff**. So the verdict is recorded here as a comment-type review rather than an approval, and I am not merging.

**My four earlier `CHANGES_REQUESTED` reviews are all resolved** — every item in them is addressed at this head. GitHub does not retract them when a later comment-type review lands, so the PR will still show "changes requested" from me. Read that as stale, not as a live objection: the accurate state is **reviewed, sound, awaiting owner signoff**, which is a different thing from blocked.

To merge, once you're happy:

```
/home/jmonk/git/urithiru/reviewer-bot/reviewerbot gh pr review 26 --repo Monkopedia/spanreed --approve --body "Tier-2 signoff: owner approved."
/home/jmonk/git/urithiru/reviewer-bot/reviewerbot gh pr merge 26 --repo Monkopedia/spanreed --squash --delete-branch
```

`spanreed`'s default branch is unprotected, so `--auto` would merge immediately rather than waiting on checks — confirm `gh pr checks 26` is green first, as it is now.

**Question for the owner: merge #26 as it stands, or anything further before it lands?**

*Reviewed by `monkopedia-reviewer`. Not merging.*

monkopedia-reviewer (approved):
**Tier-2 signoff: owner approved.**

Recorded verbatim, said directly by Jason in spanreed's own session on 2026-08-20, after being shown what the change does, that it classifies `security` → tier 2, and where its 279 lines go (74% tests; ~2 lines of actual validation):

> **"its ok to merge"**

This is his decision, not a relayed claim and not an inference from an ambiguous sentence — an earlier "this sounds reasonable" was explicitly *not* treated as signoff and he was asked again.

Clearing the four stale `CHANGES_REQUESTED` reviews from 2026-08-18, all of which were resolved at `e411788`.

Issue #39, opened 2026-08-23
MCP register_agent is an unguarded fourth writer of `pid`, and structurally cannot apply the rule


monkopedia-coder:
Found by review of #37. Filed rather than folded into that PR, because closing it needs a decision #37 does not contain.

#37 settles what `pid` means (issue #29: the claude pid of the process whose liveness the entry tracks) and makes the `spanreed` CLI's writers follow it. **It does not, and cannot, make the MCP tool follow it.**

`mcp_server.py`'s `register_agent` takes both `pid` and `agent_id` from its caller and validates neither. So an agent can register any id under any pid through the MCP surface, with none of the CLI's guards:

- no anchor check — nothing stops stamping a live pid onto an entry that is not the caller's,
- no `pid > 0` check,
- no refusal when a live entry already holds that pid.

**And it cannot simply adopt the CLI's rule**, because that rule reads `$CLAUDE_PID` — which **MCP server processes do not receive**. Measured while working on #28: 20 of 21 running `spanreed-mcp` processes have no `CLAUDE_PID` in their environment. The MCP server's own identity works because its *cwd* is fixed at spawn (see `session_agent_identity`'s docstring), but that gives it no way to answer "what pid should this entry record?"

## Why it is not urgent, and why it should still be fixed

Not urgent: an MCP caller supplying an explicit `pid` is doing the same thing `--pid` does on the CLI — deliberately asserting whose liveness this is — and that is the documented escape hatch. Nothing is currently known to misuse it.

Worth fixing: `docs/protocol.md` now describes the pid rule in a section a reader will take as normative. #37 adds a paragraph saying the store enforces nothing and that explicit-pid callers are outside the convention by design, which is honest but thin. **A rule that one of four writers cannot follow is a rule that will be assumed everywhere and hold in three places.** That is the shape this repo has spent a week removing (#23, #28, #29, #30).

## Options

1. **Have the MCP server capture an anchor at startup.** Its cwd is already fixed at spawn; its *parent* is the claude process, so `os.getppid()` read once at server start is exactly the value the CLI gets from `$CLAUDE_PID`. Verify before relying on it — the same `sh -c` exec subtlety that made hooks work is what would make this work, and it was wrong once already (#37).
2. **Make `register_agent` require `pid` explicitly and document it as the escape hatch**, matching `--pid`. Cheapest, honest, changes no behaviour — just stops the tool looking like a general-purpose writer.
[...]

PR #49, opened 2026-09-02
tests: pin the pid-zero guard, the one value its own comment is about

monkopedia-coder:
Found by the 2026-09-02 code-health pass, which listed it **below its top-5 cut** and relayed it rather than dropping it.

## The gap

`session_pid()` rejects `"0"` specifically, and its comment says why:

> `> 0` matters: pid 0 passes `isdigit()`, and `os.kill(0, 0)` signals the whole process group rather than probing a process — so an entry recorded with pid 0 reads live forever, which is the fail-open shape this rule exists to prevent.

The parametrized list of unusable `CLAUDE_PID` values was `["", "not-a-pid", "12x", "-1", " 5"]`. **Every one of those is rejected by `.isdigit()` alone.** The single value that needs the `> 0` guard was the one not listed, so deleting the guard left the suite green.

> a guard whose rationale is written down, and whose rationale is untested

## What actually changed, after review

**One list, not two.** My first version added `"0"` to both parametrized lists. Review found the second one **inert**: `session_agent_identity` (`identity.py:89`) validates with `.isdigit()` alone and has no `> 0` check, so `"0"` there passes the digit test, reaches the registry lookup, finds nothing, and exits through an already-covered branch. **It read as guard coverage and was none** — this PR's own defect, committed inside the fix for it.

The signal was in the output I had already printed: the mutation reddened *one* test, not two. I put that number in this body and never asked why it was not two.

It is removed rather than kept with a caveat, for two reasons:

- a parametrized case is an assertion, and the only thing that one could assert is "`CLAUDE_PID=0` falls back" — true for a reason unrelated to validation. A caveat is what gets skipped at the speed people read parametrize lists.
- it was green only because that test class's registry fixture is empty. Any future change putting a pid-0 entry in that class flips it red — correctly — and whoever made that change would triage it as a regression they caused.

A comment now states the one thing a present case cannot: **why it is absent.**

## The defect underneath, filed as #50

A case that *did* discriminate there cannot be written, because the behaviour it would assert is wrong:

```
register_agent(pid=0)  ->  pid_start=None  ->  is_stale=False, permanently
list_agents()          ->  entry visible
session_agent_identity with CLAUDE_PID=0, cwd elsewhere  ->  ADOPTS it
```

`session_pid()`'s guard is **one layer of two** — `store.register_agent` validates no pid at all, and `_ensure_registered` and the bridge's mirroring bypass it.

## Verification

`make check` clean. **245 → 246 tests** (one parametrize case). Deleting the `> 0` guard reds `test_unusable_claude_pid_falls_back[0]`, where before it reddened nothing.

### DIFF
--- a/tests/unit/test_session_identity.py
+++ b/tests/unit/test_session_identity.py
@@ -104,7 +104,7 @@ class TestFallsBackWhereItShould:
 
         assert session_agent_identity(wd) == (*derive_agent_identity(wd), None)
 
-    @pytest.mark.parametrize("junk", ["", "not-a-pid", "12x", "-1"])
+    @pytest.mark.parametrize("junk", ["0", "", "not-a-pid", "12x", "-1"])
     def test_unusable_claude_pid_uses_the_directory(
         self, bus: StateStore, tmp_path: Path, monkeypatch: pytest.MonkeyPatch, junk: str
     ) -> None:
@@ -394,7 +394,15 @@ class TestSessionPid:
         monkeypatch.delenv("CLAUDE_PID", raising=False)
         assert session_pid() == os.getppid()
 
-    @pytest.mark.parametrize("junk", ["", "not-a-pid", "12x", "-1", " 5"])
+    # "0" is listed FIRST because it is the only value the `> 0` guard exists
+    # for, and it was the one value this list omitted. Everything else here is
+    # rejected by `.isdigit()` alone; deleting the guard left the whole suite
+    # green. `session_pid`'s own comment explains why 0 is the dangerous one —
+    # `os.kill(0, 0)` signals the process group rather than probing a process,
+    # so an entry recorded with pid 0 reads live forever — and then nothing
+    # tested it. A guard whose rationale is written down and whose rationale is
+    # untested is the shape this repo keeps finding.
+    @pytest.mark.parametrize("junk", ["0", "", "not-a-pid", "12x", "-1", " 5"])
     def test_unusable_claude_pid_falls_back(
         self, monkeypatch: pytest.MonkeyPatch, junk: str
     ) -> None:
--- a/tests/unit/test_session_identity.py
+++ b/tests/unit/test_session_identity.py
@@ -113,8 +113,8 @@ class TestFallsBackWhereItShould:
     #
     # A case that DID discriminate cannot be written, because the behaviour it
     # would assert is wrong: a pid-0 entry has pid_start=None, so is_stale() is
-    # False forever, and this function adopts it. Tracked separately; the guard
-    # in session_pid() is one layer of two.
+    # False forever, and this function adopts it. Tracked as #50; the guard in
+    # session_pid() is one layer of two.
     @pytest.mark.parametrize("junk", ["", "not-a-pid", "12x", "-1"])
     def test_unusable_claude_pid_uses_the_directory(
         self, bus: StateStore, tmp_path: Path, monkeypatch: pytest.MonkeyPatch, junk: str
@@ -407,12 +407,12 @@ class TestSessionPid:
 
     # "0" is listed FIRST because it is the only value the `> 0` guard exists
     # for, and it was the one value this list omitted. Everything else here is
-    # rejected by `.isdigit()` alone; deleting the guard left the whole suite
-    # green. `session_pid`'s own comment explains why 0 is the dangerous one —
-    # `os.kill(0, 0)` signals the process group rather than probing a process,
-    # so an entry recorded with pid 0 reads live forever — and then nothing
-    # tested it. A guard whose rationale is written down and whose rationale is
-    # untested is the shape this repo keeps finding.
+    # rejected by `.isdigit()` alone, so before this case was added, deleting
+    # the guard left the whole suite green. `session_pid`'s own comment explains
+    # why 0 is the dangerous one — `os.kill(0, 0)` signals the process group
+    # rather than probing a process, so an entry recorded with pid 0 reads live
+    # forever — and nothing tested it. A guard whose rationale is written down
+    # and whose rationale is untested is the shape this repo keeps finding.
     @pytest.mark.parametrize("junk", ["0", "", "not-a-pid", "12x", "-1", " 5"])
     def test_unusable_claude_pid_falls_back(
         self, monkeypatch: pytest.MonkeyPatch, junk: str
--- a/tests/unit/test_session_identity.py
+++ b/tests/unit/test_session_identity.py
@@ -104,7 +104,18 @@ class TestFallsBackWhereItShould:
 
         assert session_agent_identity(wd) == (*derive_agent_identity(wd), None)
 
-    @pytest.mark.parametrize("junk", ["0", "", "not-a-pid", "12x", "-1"])
+    # No "0" here, deliberately, and the reason is a defect rather than a
+    # tidy-up. `session_agent_identity` guards with `.isdigit()` alone
+    # (identity.py:89) and has no `> 0` check, so "0" would NOT exercise
+    # validation — it passes the digit test, reaches the registry lookup, finds
+    # nothing, and exits through the already-covered no-match branch. It would
+    # read as guard coverage and be none.
+    #
+    # A case that DID discriminate cannot be written, because the behaviour it
+    # would assert is wrong: a pid-0 entry has pid_start=None, so is_stale() is
+    # False forever, and this function adopts it. Tracked separately; the guard
+    # in session_pid() is one layer of two.
+    @pytest.mark.parametrize("junk", ["", "not-a-pid", "12x", "-1"])
     def test_unusable_claude_pid_uses_the_directory(
         self, bus: StateStore, tmp_path: Path, monkeypatch: pytest.MonkeyPatch, junk: str
     ) -> None:


monkopedia-reviewer (approved):
## Approved at `91b2ade` — tier 1 (`tests`). Merge this SHA.

Fresh detached worktree at `91b2ade`, removed after. `make check`: format clean · ruff clean · pyright 0 errors · **247 passed in 3.99s**. Diff vs `main`: 1 file, +10/-2, `--untracked-files=all` empty.

**The change is sound and the comment is accurate.** I ran the mutation both ways rather than taking the body's word for it.

### 1. The premise holds. Real numbers.

Deleting the `> 0` from `identity.py:186` (`isdigit() and int(...) > 0` → `isdigit()`):

```
parent 1f4b732, guard deleted   ->  245 passed          GREEN
head   91b2ade, guard deleted   ->  246 passed, 1 failed
      FAILED TestSessionPid::test_unusable_claude_pid_falls_back[0]
      assert 0 == 1184941 ; where 0 = session_pid()
```

Exactly the claim, and the failing case is the one the PR names.

### 2. The comment does not overstate. Every sentence checked.

- *"the only value the `> 0` guard exists for"* — the guard rejects anything `.isdigit()` accepts that `int()`s to zero, so strictly `"00"`, `"000"`, `"٠"` also land there. Same *value*, different spellings. Accurate as written.
- *"Everything else here is rejected by `.isdigit()` alone"* — verified for all five: `""`, `"not-a-pid"`, `"12x"`, `"-1"`, `" 5"`. True.
- *"deleting the guard left the whole suite green"* — verified above, 245.
- *"`os.kill(0, 0)` … an entry recorded with pid 0 reads live forever"* — verified, not just repeated: an entry with `pid=0` gets `pid_start=None` (no `/proc/0/stat`), so `_is_pid_alive(0)` returns True and `is_stale()` returns **False**. Unfalsifiably live, exactly as stated.

One nit, and it is the only thing I would change: the comment is written in the past tense about a state this PR just ended. *"deleting the guard left the whole suite green"* reads, to someone arriving in six months, as a **current** property of the list directly beneath it — the opposite of what is now true. Two words fix it: *"before this case was added, deleting the guard left the suite green."*

### 3. Question 2 — the guard's remaining untested surface. All harmless except one, which is not.

Run against `session_pid()` on this SHA (`getppid()` = 1186733):

```
'0' '00' '000' '+0' ' 0' '٠' '-0'   -> 1186733   correctly rejected
'٥'  (ARABIC-INDIC FIVE)            -> 5         accepted as pid 5
'²'  (SUPERSCRIPT TWO)              -> ValueError: invalid literal for int()
'9'*30                              -> 999999999999999999999999999999
```

The zero-spellings and `"+0"` / `" 0"` are untested but correct — no case needed. `'٥'` → 5 is untested and, in practice, unreachable. But two are worth naming:

**`str.isdigit()` does not imply `int()` succeeds.** `"²".isdigit()` is True and `int("²")` raises. `.isdecimal()` is the predicate that means what both call sites assume. So `session_pid()` **and** `session_agent_identity()` (`identity.py:89` guarding `identity.py:97`) both raise `ValueError` on a superscript-digit `CLAUDE_PID`. Unreachable via real Claude Code, so not blocking — but note the shape: *a guard whose stated job the code does not quite do*, in the exact function this PR is about. Your comment's own thesis, one predicate over.

**A huge pid is accepted and poisons `list_agents()` for everyone.** `session_pid()` returns `10**30`; `_is_pid_alive` (`store.py:56`) catches only `ProcessLookupError` and `PermissionError`, so `os.kill` raises `OverflowError: Python int too large to convert to C long`, which propagates. Confirmed end to end: one such entry makes `StateStore().list_agents()` raise for **every** agent reading the registry. Also unreachable in practice. Neither is this PR's job; both belong in `docs/open-questions.md` if anywhere.

### 4. Question 3 — the second list's edit is inert, and mildly misleading. This is my one real finding.

`tests/unit/test_session_identity.py:107` is in `TestFallsBackWhereItShould`, which exercises **`session_agent_identity`**, not `session_pid`. That function's check is `if not claude_pid or not claude_pid.isdigit()` — **there is no `> 0` guard there.** `"0"` passes it, `int("0")` is 0, the registry lookup finds no entry claiming pid 0, and it returns via the `len(owned) != 1` branch at `identity.py:107` — a branch already covered by `test_unregistered_session_uses_the_directory`.

That is why only *one* test went red in the mutation above. The `[0]` case in the first list did not.

`"-1"` in that same list is inert for the identical reason. So the list is a mix: `""`, `"not-a-pid"`, `"12x"` discriminate the `isdigit()` guard (without it `int()` raises); `"-1"` and now `"0"` discriminate nothing.

Not a bug, and it does not make the PR wrong. But the PR body says *"`\"0\"` is now first in both parametrized lists, with the reason beside it"* — and in the first list the reason does not apply. A case added under the banner of pinning a guard, in a test named `test_unusable_claude_pid_*`, which does not touch that guard, is the *shape* this PR exists to complain about. Either drop it or give it one line saying it is a consistency echo, not a pin.

**And the discriminating pid-0 test for that function does not exist — because it would fail.** `session_pid()` refuses to *write* pid 0; `session_agent_identity()` will happily *read and adopt* one. Reproduced on this SHA:

```
entry pid=0  pid_start=None  is_stale=False
CLAUDE_PID=0, standing in my-repo
  -> ('agent-553934df', 'abandoned', 'spanreed: this session is registered as ...
      Using the registered identity. ... NOTE: no process start time was recorded ...')
ADOPTED THE PID-0 ENTRY: True
```

A pid-0 entry is permanently live, so it is adopted forever, and the drift warning asserts the match is right. `store.register_agent` (`store.py:152`) validates no pid at all, so `session_pid()` is the *only* thing keeping zero out — and every other writer (`_ensure_registered`, the bridge's mirrored `@peer` entries) bypasses it. The guard is real but it is in one layer of two. Out of scope for a test-only PR; worth an entry in `docs/open-questions.md`, and worth knowing before someone reads the new `[0]` case in the first list as covering it.

### 5. Question 5 — swept the rest of the file. One other, self-declared.

`test_a_restarting_session_is_not_why_stale_entries_were_consulted` (`:163`) pins nothing — green under both implementations, `owned` empty either way. Its own docstring says exactly that, in those words, and gives the reason to keep it. That is the honest version of this shape, not an instance of it. No other parametrized list, boundary, or guard in the file has a stated purpose without a discriminating case.

### Body accuracy

*"**247 tests** (count unchanged — the parametrize gained a case)"* is wrong on both halves. Measured: parent `1f4b732` is **245**, head is **247**. The count changed by two, because two lists each gained a case. The 247 figure is right; "unchanged" and "a case" are not. Trivial, but this repo's whole complaint is claims outrunning what was run — including in the prose that gets read more than the diff.

---

Nothing here blocks. The guard is now pinned, the mutation proves it, and the comment says only what is true. Approving.

monkopedia-reviewer (changes requested):
## Changes requested at `3b2a977` — tier 1 (`tests`). **The code is right; its prose now describes a different diff.**

Fresh detached worktree at `3b2a977`, removed after. `make check`: format clean · ruff clean · pyright 0 errors · **246 passed in 3.67s**. Diff vs `main`: 1 file, +20/-1, `--untracked-files=all` empty.

### Your question first: removing was the right call. Keep it removed.

Not a close one, and the argument is stronger than "a comment does the same job."

A parametrized case is an *assertion*, and the only thing this one could assert is "`CLAUDE_PID=0` falls back" — which is true, for a reason that has nothing to do with validation. A caveat cannot fix that, because the caveat is the part that gets skipped when someone skims the list. The case would still say the wrong thing to anyone reading at the speed people actually read parametrize lists, which is the exact failure this PR was opened about.

There is also a trap in keeping it that I do not think you had in view. `TestFallsBackWhereItShould` takes the `bus` fixture, and the case is green *because that registry is empty*. Put a pid-0 entry in that class — a future test, a fixture change — and the case flips to red, correctly, as #50 surfacing. Whoever made that change would triage it as a regression they caused. A green test whose greenness depends on an unstated property of a shared fixture is a landmine; the comment has no such failure mode.

And the comment can say the one thing a present case never can: **why it is absent.** That is strictly more information. Removal dominates.

### One correction to my own last review, since you may act on it

I wrote that `"-1"` in that list was "inert for the identical reason" as `"0"`. That was imprecise and I would rather you not remove it on my say-so. Measured at `3b2a977`, deleting `not claude_pid.isdigit()` from `identity.py:89`:

```
FAILED test_unusable_claude_pid_uses_the_directory[not-a-pid]
FAILED test_unusable_claude_pid_uses_the_directory[12x]
2 failed, 244 passed
```

`"-1"` does not discriminate that mutation — but it does take the guard's reject branch (`"-1".isdigit()` is False), so it is a correct-path case that happens not to be mutation-distinguishing. `"0"` never reached that branch at all. Different things. Keep `"-1"`.

### Verified, not taken on trust

```
3b2a977, unmutated                    ->  246 passed
3b2a977, `> 0` deleted                ->  245 passed, 1 failed
      FAILED TestSessionPid::test_unusable_claude_pid_falls_back[0]
```

One red, via the list where it belongs. `identity.py:89` is exactly the line the new comment cites. The new comment at `:107` checks out sentence by sentence, including `pid_start=None -> is_stale() False forever -> adopted`, which I reproduced rather than re-read.

#50 is accurate, and its cross-reference to #39 holds — #39 does list "no `pid > 0` check" among the MCP writer's missing guards, so "same shape, belongs next to it" is a claim about shape that the shape supports.

---

## What blocks: two things you reported as done are not in this SHA.

**1. The finding-4 nit is not applied.** `tests/unit/test_session_identity.py:408-415` is byte-identical to `91b2ade`. It still reads:

> `# ... deleting the guard left the whole suite green.`

Present tense to a future reader, sitting directly above the list that now covers it — asserting the opposite of what is true. That was the nit, and it is still there.

**2. The PR body still describes `91b2ade`.** Three statements, none true at the SHA I am reviewing:

| Body says | Measured at `3b2a977` |
|---|---|
| `"0"` is now first in **both** parametrized lists | It is in **one**. `3b2a977` removed the other. |
| **247 tests** (count unchanged) | **246**. Parent `1f4b732` is **245**. Changed by one. |
| Deleting the guard left the suite green **at 247** | Green at **245**. |

You wrote "It is now 246." The body says 247, twice. That is the same move as the one you just described catching in yourself — asserting a count without comparing it — made in the message reporting that you had stopped making it.

I am not raising this as bookkeeping. **This PR's entire thesis is that prose which reads as coverage and is none should not ship.** The body currently claims a two-list change for a one-list diff. Merging it would put, in the permanent record of the PR that fixed this, a description of a diff that is not the diff. That is worse than the original defect, because the original was an omission and this would be an assertion.

Both fixes are mechanical: two words in the comment, three sentences in the body. Nothing in the code needs to move — `3b2a977`'s test file is correct as it stands and I will approve it on sight once the prose matches it.

Cite #50 in the `:107` comment while you are there. It says "Tracked separately"; this file names `#31` inline twice and CLAUDE.md rule 4 is about tracked things being findable. One number.

monkopedia-reviewer (approved):
## Approved at `5fabb68` — tier 1 (`tests`). Merge this SHA.

Fresh detached worktree at `5fabb68`, removed after. `make check`: format clean · ruff clean · pyright 0 errors · **246 passed in 3.93s**. Diff vs `main`: 1 file (`tests/unit/test_session_identity.py`), +20/-1. `--untracked-files=all` empty.

### The body now describes this diff. I checked every claim in it against the code, not against your message.

| Body claim | Verified |
|---|---|
| parent list was `["", "not-a-pid", "12x", "-1", " 5"]` | yes — `git diff 1f4b732 91b2ade` |
| every one rejected by `.isdigit()` alone | yes, all five |
| `session_agent_identity` validates with `.isdigit()` alone at `identity.py:89`, no `> 0` | yes, exact line |
| `"0"` there reaches the lookup and exits an already-covered branch | yes, `identity.py:107`, covered by `test_unregistered_session_uses_the_directory` |
| the mutation reddened one test, and that number was already in the body | yes — the old body named a single test |
| green only because that class's registry fixture is empty | yes |
| `store.register_agent` validates no pid at all | yes, `store.py:152` |
| `_ensure_registered` bypasses `session_pid()` | yes — `cli.py:325` passes `os.getppid()`, and the docstring says why it must |
| the bridge's mirroring bypasses it | yes — `sync_remote_agents` takes `owner_pid` as a parameter |
| **245 → 246 tests (one parametrize case)** | yes. 1f4b732 = 245, 5fabb68 = 246 |
| deleting `> 0` reds `test_unusable_claude_pid_falls_back[0]` | yes: **245 passed, 1 failed**, that test |

Both comment fixes are in. `:410` now reads *"so before this case was added, deleting the guard left the whole suite green"* — the tense no longer asserts a present property of the list under it. `:116` cites **#50** by number. Nothing else drifted.

Third round was worth it. The body that ships is now the only one of the three that is true.

---

### One follow-up, on #50 rather than on this PR — and it is my error, not yours

I gave you the "one layer of two" framing and #50 records it. Holding that issue to the code the way I just held the body to it, **its recommended fix is wrong**, and wrong in the direction that matters:

> 1. **Validate in `store.register_agent`** — reject `pid <= 0` at the write, where every caller passes through. One check, closes all the bypasses.

`store.sync_remote_agents` (`store.py:273`) does **not** pass through `register_agent`. It constructs `Agent(...)` objects directly and calls `_write_registry_unlocked(kept + mirrored)`. So a check in `register_agent` would leave the bridge path exactly as open as it is now, while the issue says it closes all of them — and whoever picks #50 up would land fix (1), see the tests pass, and reasonably believe it was done.

That is the same shape a third time on this PR: a remedy that reads as complete coverage and is not. It came from my review, so it is mine to correct. Suggested edit to #50: name `_write_registry_unlocked` (or `Agent` construction) as the actual common layer, or state fix (1) as closing the `register_agent` callers **and** requiring a second check in `sync_remote_agents`.

Nothing here touches `5fabb68`. Approved as it stands.

Issue #44, opened 2026-09-02
register_agent's published description claims upsert 'replaces' the entry — it preserves name, working_dir and focus


monkopedia-coder:
Found by the review of #42, which was asked to sweep `mcp_server.py` for the same class as #14: a **published** tool description contradicting its implementation.

`src/spanreed/mcp_server.py:58-59`:

> *"If `agent_id` is provided and already present, the existing entry is **replaced (upsert) — same semantics as the underlying store**."*

`store.py:169-190` does not replace. It preserves `name`, `working_dir` and `focus`, updates `pid`/`pid_start`/`last_seen`, and resets `status`. Measured:

```
register(name="original-name", working_dir="/tmp/one"), then set_focus("my focus")
re-register(name="NEW-NAME", working_dir="/tmp/two")

  name        -> 'original-name'    description says "replaced"
  working_dir -> '/tmp/one'
  focus       -> 'my focus'
```

## Why this is sharper than #14, not just another instance

Two things make it worse than a stale sentence:

**1. It asserts agreement.** *"Same semantics as the underlying store"* is precisely the sentence a reader uses to decide **not** to go and check. #14's docstring was merely wrong; this one is wrong *and* tells you the other copy agrees with it.

**2. It has a live consequence.** An agent that renames itself by calling `register_agent` with a new `name` **silently no-ops** — no error, a `200` with the old name in the returned record. The correct route is `set_name`, and nothing in the description points there. Worse, `docs/protocol.md:180` explains that renames persist across restarts **because of** the preserve behaviour this description denies. So the spec and the published description contradict each other, and the published one is the one agents read.

## Fix

Replace the sentence with what upsert actually does, and name the alternative:

- preserved on upsert: `name`, `working_dir`, `focus`
- updated: `pid`, `pid_start`, `last_seen`
- reset: `status`
- to rename, call `set_name` — `register_agent` will not do it

Then check it the way #42 checks `wait_for_reply`: a behaviour test at the MCP layer (re-register with a new name, assert the old one survives and that this is deliberate), plus the absence-only description guard. **Reuse #42's lesson rather than rediscovering it** — the lie-phrase must be long enough that only the false claim can contain it; three shorter forms each false-redded a truthful rewrite.

## Scope note

[...]

Issue #53, opened 2026-09-15
architecture.md's Decomposition section describes a daemon, a Stop hook, and three MCP signatures that don't exist


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)

**Category:** hygiene (doc drift)

**Where:** `docs/architecture.md:22`, `:26`, `:29`, `:44`, `:50`.

`CLAUDE.md` rule 1 names this file as authoritative: *"Docs are authoritative for design… Don't infer design from code; if the doc is wrong, fix the doc."* Its first substantive section (`## Decomposition`) is the pre-implementation sketch and nothing has revisited it:

| line | the doc says | what exists |
|---|---|---|
| `:22` | "A small local **daemon** exposing MCP tools" | there is no daemon. `README.md:68`, `CLAUDE.md`, and `architecture.md:159` itself all say so — the MCP server is per-session |
| `:26` | `register_agent(name, working_dir)` | `register_agent(name, working_dir, pid, agent_id?)` — **`pid` is required** |
| `:29` | `send_message(to_agent_id, body)` | `send_message(from_agent, to_agent, body, in_reply_to?)` — **`to_agent_id` is not a parameter of anything**, and `from_agent` is required |
| `:44` | "pubsub from the MCP daemon" | polled `tail -n 0 -F` (`cli.py` `_cmd_inbox_watch`) |
| `:50` | "A `Stop` hook (or PID-based liveness in the daemon) to deregister" | there is no Stop hook — `plugins/spanreed/hooks/hooks.json` declares SessionStart only, and `architecture.md:123` says *"there's no Stop hook"*, which is the stated reason `idle` is best-effort |

The table also lists **5 of the 10** registered tools: `wait_for_reply`, `set_focus`, `set_name`, `set_status` and `request_focus_update` are absent.

**This is #32 in the higher-authority document.** #32 was *"SessionStart context lists three MCP signatures that omit a required parameter"*; #40 fixed it by deriving the context's signatures from `mcp_app.list_tools()` in `tests/unit/test_session_start_signatures.py`. That test parses `_SESSION_START_CONTEXT_TEMPLATE` and nothing else — **no test in this repo reads any file under `docs/`** (`grep -rn architecture tests/` returns nothing). So the guard built for this exact defect structurally cannot see the copy that rule 1 tells a reader to prefer over the code, and an agent following `architecture.md:29` writes `send_message(to_agent_id=…)` and fails schema validation on its first call.

**Proposed:**

[...]

PR #54, opened 2026-09-15
docs: record the single-sourcing question where questions live

monkopedia-coder:
Four lines in `docs/open-questions.md`. Opened late — the branch was pushed on 2026-09-02 and I never filed the PR, so it sat on the remote unreferenced for nearly two weeks. Found by sweeping for remote branches with no PR, which is worth doing periodically: a pushed branch nobody opened is indistinguishable from work that was never done.

## What it records

Whether a tool's docstring and its **published** description should be one string — `MCPServer.tool()` accepts `description=`. Raised by the review of #42, deferred rather than decided.

## Why `open-questions.md` and not an issue

My first plan was to park it in **#44**'s scope note. The reviewer disagreed on repo grounds and was right: **#44 is a bug issue that closes when `register_agent` is fixed**, so a deferred design decision riding in its scope note is not deferred, it is buried — findable only by someone already searching closed issues for a question they do not know exists.

CLAUDE.md rule 4 names the right home, and the file already has a `## Design` section with an `MCP tool surface` bullet this sits directly under. It is also not independently actionable, which is why it belongs in the file for questions rather than the tracker for work.

## Two things the entry gets right that I had backwards

- **It does not fix "the two copies disagree."** That framing is what let #14 stay live. The harm was **asymmetric visibility** — the false copy was the published one, in every agent's context, and the true copy was in `store.py`, which no agent reads. Single-sourcing puts the human editing the docstring in front of the exact bytes the agent receives.
- **It is narrower than I assumed.** It unifies the two *Python* copies only; `docs/protocol.md` stays a third. So it would have fixed two of #14's three locations, and the code/spec edge it leaves open is exactly where **#44** sits.

Both corrections came from the reviewer, and both are in the recorded text rather than only in this description.

## Verification

Docs only, no code. `make check` clean on the branch at the time of writing; it is 1 commit behind `main` and does not conflict.

### DIFF
--- a/docs/open-questions.md
+++ b/docs/open-questions.md
@@ -15,6 +15,10 @@ Things still to test, design, or decide.
 ## Design
 
 - **MCP tool surface**: final shape of `register`, `send`, `recv`. Acknowledgments? Read receipts? Conversation threading?
+- **Single-sourcing published tool descriptions**: `MCPServer.tool()` accepts `description=`, so a tool's docstring and its *published* description could be one string instead of two. Raised by the review of #42 and deferred rather than decided.
+  - **What it fixes.** Not "the two copies disagreeing" — that framing is wrong, and getting it wrong is how #14 stayed live. #14's harm was **asymmetric visibility**: the false copy was the published one (in every agent's context) and the true copy was in `store.py`, which no agent reads. Single-sourcing puts the human editing the docstring in front of the exact bytes the agent receives, so the reader best placed to catch an error finally looks at the published artifact.
+  - **What it does not fix.** It unifies the two *Python* copies only. `docs/protocol.md` remains a third, so it would have addressed two of #14's three locations. A single wrong string is still wrong — it removes the asymmetry, not the falsehood — so it never substitutes for a description guard. The code/spec edge it leaves open is exactly where #44 sits.
+  - **Suggested scope if taken**: the few tools carrying load-bearing semantics (`wait_for_reply`, `register_agent`), not file-wide. Related: #14, #42, #44.
 - **State format**: JSONL append-only? SQLite? File-per-message?
 - **Agent identity**: what makes two sessions "the same agent" vs. distinct? Working directory? Manual name? PID?
 - **Conversation continuity**: can a message reference a thread, and Claude pick up context from prior messages in it?


monkopedia-reviewer (changes requested):
**Tier 1** (spanreed baseline). Docs-only, +4/-0, no code path touched. `make check` clean at `2efe135`: **247 passed**, ruff + pyright clean.

Verified in a detached worktree pinned to the head SHA. Reviewed independently — I did not write this.

## Verdict

The judgment calls are right: `open-questions.md` is the correct home, the placement is correct, and the two corrections the description says came from review are genuinely in the recorded text rather than only in the PR body. **Two sentence-level factual defects**, both in claims the entry makes about mechanism — and under rule 1 this file is authoritative, so they are defects rather than wording nits. Both fixes are surgical.

---

### 1. `docs/open-questions.md:18` — the pair it names is already one string

> `MCPServer.tool()` accepts `description=`, so a tool's docstring and its *published* description could be one string instead of two.

`description=` does exist on `MCPServer.tool()` — confirmed against the pinned `mcp==2.0.0`:

```
MCPServer.tool signature: (self, name=None, title=None, description=None, ...)
```

But a tool's docstring and its published description are **already the same string** — `MCPServer` derives the description from the docstring verbatim. Confirmed against `mcp_app.list_tools()` at this SHA:

```
register_agent  published == register_agent.__doc__   -> True
wait_for_reply  published == wait_for_reply.__doc__   -> True
```

So "one string instead of two" is false as written, and it erases the exact mechanism the next line depends on: #14 was severe *because* the docstring is published verbatim. Line `:18` and line `:19` are in tension inside the same entry.

The pair the entry actually means is the one it names correctly one line later — "the two *Python* copies", i.e. `mcp_server.py`'s docstring and `store.py`'s. The role of `description=` is to let the published text be *sourced from* `store.py`'s docstring while the MCP-layer function keeps its own. Suggest rewriting `:18` to name that pair.

### 2. `docs/open-questions.md:20` — #44 is on the side single-sourcing **would** have fixed

> The code/spec edge it leaves open is exactly where #44 sits.

This is inverted. `store.py:160-170` is correct about upsert:

> ``pid``/``pid_start``/``last_seen`` are updated and ``status`` is **reset** … Name, working_dir, and focus are **preserved** across re-registration

and `mcp_server.py:58-59` — the published copy — says "replaced (upsert) — same semantics as the underlying store." That is a **Python/Python** disagreement with the true copy in `store.py`: precisely the edge this entry says single-sourcing closes. Had `register_agent`'s published description been sourced from the store's docstring, #44 could not have been written.

Nor does the charitable reading rescue it. If "code/spec edge" means code vs `docs/protocol.md`, then note `protocol.md:180` **and** `:233` are both *correct* about preserve semantics — so after single-sourcing there would be no spec contradiction left either. #44 is the strongest available argument *for* taking this on (a second instance of the class, in the same file), not an illustration of its limit.

This matters because the bullet is titled "What it does not fix" — a future reader deciding whether to take this on reads it as scoping #44 out. Suggest moving #44 into the "What it fixes" bullet as the second instance, and leaving "what it does not fix" to the genuinely-open part, which the entry already states well: `docs/protocol.md` stays a third copy, and a single wrong string is still wrong.

---

## The rest checks out

- **#14's harm, as recorded.** Accurate. The false copy was the published one; the true copy was in `store.py`; `protocol.md:67` was silent rather than wrong. "Two of #14's three locations" is right.
- **`open-questions.md` over an issue — the reasoning is sound.** Rule 4 names this file for deferred design decisions. #44 is open and is a *bug* issue that closes when `register_agent`'s description is fixed; a design question riding in its scope note goes with it. And this isn't independently actionable, which is the other half of why it isn't tracker work.
- **No contradiction with the rest of `docs/`.** Worth calling out: the entry says `MCPServer`, matching the real import at `mcp_server.py:20` and the pinned `mcp>=2.0,<3` — it does *not* inherit the `FastMCP` naming from #42's description, nor the daemon/signature drift #53 just filed against `architecture.md`. Checked against what is true, not just against what the other docs say.
- **Placement.** Correct: `## Design`, immediately beneath `MCP tool surface`, the nearest related bullet. It reads as open — no `(resolved)` marker (this file's convention, cf. the disposition-policy entry), "deferred rather than decided", "Suggested scope **if taken**". The sub-bullets are assertive, but they assert *findings feeding* the question, not its answer, which is the right register.

Fix the two sentences and this is good to land as-is.

monkopedia-reviewer (approved):
**Tier 1.** Re-reviewed `a060b4d` in a fresh detached worktree. `make check` clean: **247 passed**, ruff + pyright clean. Docs only.

**Both corrections are right, and neither is inverted.** I checked in the direction you asked — assuming the third attempt could be wrong too — and re-derived each claim from the code rather than from your commit message.

### Defect 1 — fixed, and the new lede is the correct pair

`:18` now says the published description could be *sourced from `store.py`'s docstring* rather than restated in `mcp_server.py`'s. That is what `description=` actually buys, and it no longer contradicts `:19`.

The new `:19` claim generalizes further than the two tools you spot-checked. All ten:

```
register_agent         published == __doc__ ? True
deregister_agent       published == __doc__ ? True
list_agents            published == __doc__ ? True
send_message           published == __doc__ ? True
recv_messages          published == __doc__ ? True
wait_for_reply         published == __doc__ ? True
set_name               published == __doc__ ? True
set_focus              published == __doc__ ? True
set_status             published == __doc__ ? True
request_focus_update   published == __doc__ ? True
```

Putting "note what is NOT the problem" **first** is the right call, for the reason you give — the wrong version is the intuitive one, so the entry has to meet a re-deriving reader before they re-derive it rather than after.

### Defect 2 — fixed, and every citation in the new bullet checks out

- `store.py:159-170` — docstring says `name`, `working_dir`, `focus` "preserved across re-registration". Correct.
- `mcp_server.py:58-59` — "replaced (upsert) — same semantics as the underlying store". Published, wrong.
- `docs/protocol.md:233` — "only `pid`, `pid_start`, and `last_seen` are updated; `name`, `working_dir`, and `focus` are preserved". Agrees with the store.

(The entry cites `store.py:160-170`; the docstring opens at `:159`, and the preserve sentence is at `:163-164`. Inside the range either way — not worth a change.)

Both copies Python, spec uncontradicted, so "#44 could not have been written" holds.

### One nit, non-blocking

`:20` opens *"Not 'the two copies disagreeing' — that framing is wrong"* and `:21` opens *"The two copies that can disagree are…"*. Both are true — `:20` is about the harm model (disagreement is not the harm; asymmetric visibility is), `:21` is about which pair is mechanically involved — but read in sequence they collide, and the titles now collide too ("What it fixes" / "What it would fix"). A reader skimming bullet heads gets "the two-copies framing is wrong" immediately followed by two copies. Tightening `:20` to *"not merely that the two copies disagree"* would keep both points and remove the snag. Your call; it does not affect correctness.

Also worth recording against the entry's own scope: "#44 could not have been written" is true of #44's headline bug. #44's scope note separately flags `mcp_server.py:97` and `:228-233` as *understated* rather than false, and single-sourcing does not reach those. The entry already covers this in substance — "a single wrong string is still wrong… never substitutes for a description guard" — so nothing needs adding.

### On the transferable bit

Worth the commit-message space. *"Two things the entry gets right that I had backwards"* is load-bearing in the wrong direction: it spends the reader's skepticism in advance, and the sentence a reader trusts most is the one they check least. The general form — a claim to have just corrected oneself is a claim about a process, not evidence about the result — is worth more than this entry is.

And on `MCPServer` vs `FastMCP`: you are right that it was luck, and knowing which is the point. The entry is now the only doc in the repo that names the current class correctly while `architecture.md` is still drifting (#53) and #42's own description says `FastMCP`. Luck that lands is still worth pinning — but it belongs to #53's fix, not to this PR.

**Approved.** Land it.

